PR #4, opened 2019-02-07
Feature/suma form

arbulu89:
Upgrade salt formula to work with SUMA
- SUMA form and metadata created
- New pre-validation state added to work with SUMA data
- Common state added to install shaptools correctly

### DIFF
--- a/hana/common.sls
+++ b/hana/common.sls
@@ -1,10 +1,24 @@
-#required packages to install SAP HANA, maybe they are already installed in the
-#used SLES4SAP distros
-numactl:
-  pkg.installed
+#required packages to install SAP HANA
 
-libltdl7:
-  pkg.installed
+{% set pattern_available = 1 %}
+{% if grains['os_family'] == 'Suse' %}
+{% set pattern_available = salt['cmd.retcode']('zypper search patterns-sap-hana') %}
+{% endif %}
+
+{% if pattern_available == 0 %}
+{% set repo = salt['pkg.info_available']('patterns-sap-hana')['patterns-sap-hana']['repository'] %}
+patterns-sap-hana:
+  pkg.installed:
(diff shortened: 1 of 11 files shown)


hana/common.sls
```diff
@@ -1,15 +1,20 @@
-git-core:
-  pkg.installed:
-    - refresh: False
+#required packages to install SAP HANA, maybe they are already installed in the
```
diegoakechi: @arbulu89  SLES for SAP has a pattern with all the packages to install HANA. I think we should use it, as it is maintained and will be automatically be reflected here in case of new versions.

It is called: patterns-sap-hana

diegoakechi: Complementing, only if it is SLES for SAP, as it will not work in other distros ;)

krig:
Looks good to me!

hana/common.sls
arbulu89: @diegoakechi Good point of vew. Anyway, is this packages available only for SLES4SAP distros? Can we use it otherwise? Do we need to be connected to a suse channel?
And finally, is this packages going to install extra stuff that is not really needed? I guess so.
Even with this, do you think that we should install this way?

diegoakechi: Yes, you only have access if using the SLES 4 SAP product, but it brings an aggregated value for the users, as the pattern is constantly tested and maintained. As we have plans to review the granularity of the pattern, I think is ok to merge it and open an issue for future adjustment when we have a more granular patterns organization.

arbulu89: @diegoakechi I have updated the way we install the packages to use a pattern in the case the distro is Suse and the pattern is available. Otherwise it will try to install the minimum required packages.
What do you think/
I will update habootstrap-formula to do the same too!

PR #7, opened 2019-02-20
Add RA templates and update packaging files

arbulu89:


saphanabootstrap-formula.spec
```diff
@@ -44,13 +55,40 @@ SAP HANA deployment salt formula
 pwd
 mkdir -p %{buildroot}/srv/salt/
 cp -R %{fname} %{buildroot}/srv/salt/%{fname}
+cp -R %{ftemplates} %{buildroot}/srv/salt/%{fname}/%{ftemplates}
+
+# SUMA Specific
+mkdir -p %{buildroot}/usr/share/susemanager/formulas/states/%{fname}
```
diegoakechi: @arbulu89 I suggest to change the /usr/share to the macro %{_datadir}

saphanabootstrap-formula.spec
```diff
@@ -44,13 +55,40 @@ SAP HANA deployment salt formula
...
+mkdir -p %{buildroot}/usr/share/susemanager/formulas/metadata/%{fname}
+cp -R %{fname} %{buildroot}/usr/share/susemanager/formulas/states/%{fname}
+cp -R %{ftemplates} %{buildroot}/usr/share/susemanager/formulas/states/%{fname}/%{ftemplates}
```
diegoakechi: @arbulu89 I think it is missing the directory creation for the templates directory.

saphanabootstrap-formula.spec
```diff
@@ -44,13 +55,40 @@ SAP HANA deployment salt formula
...
+%license LICENSE
+%doc README.md
+/usr/share/susemanager/formulas/states/%{fname}
+/usr/share/susemanager/formulas/states/%{fname}/%{ftemplates}
+/usr/share/susemanager/formulas/metadata/%{fname}
+
+%dir %attr(0755, root, salt) /usr/share/susemanager/
+%dir %attr(0755, root, salt) /usr/share/susemanager/formulas/
+%dir %attr(0755, root, salt) /usr/share/susemanager/formulas/states/
+%dir %attr(0755, root, salt) /usr/share/susemanager/formulas/metadata/
```
diegoakechi: @arbulu89 Is not needed to set the attrib for the templates directory too?

hana/common.sls
```diff
@@ -20,13 +20,6 @@ install_required_packages:
 
 {% endif %}
 
-# Temporary solution to import the needed packages
-add-saphana-repo:
-  pkgrepo.managed:
-    - name: saphana
-    - baseurl: https://download.opensuse.org/repositories/home:xarbulu:sap-deployment/SLE_12_SP4/
```
diegoakechi: @arbulu89 This is still needed, but the should be changed to the new repo: https://download.opensuse.org/repositories/network:/ha-clustering:/Factory/SLE_12_SP4/

saphanabootstrap-formula.spec
arbulu89: Ok! I will copy that from habootstrap-formula

saphanabootstrap-formula.spec
arbulu89: @diegoakechi I don't think this is really needed. The recursive copy creates the folder too

saphanabootstrap-formula.spec
arbulu89: @diegoakechi I have tested the installation and it worked fine. But i will double check!

hana/common.sls
arbulu89: @diegoakechi As discussed, the repositories must be added before using the formula as prerequisite

diegoakechi: OK. It makes the formula clean.

PR #9, opened 2019-02-22
Adjust the labels to SAP terminology and include help texts.

diegoakechi:


form.yml
```diff
@@ -145,17 +158,19 @@ hana:
...
         remote_host:
           $name: Primary node hostname
           $type: text
           $optional: false
+          $help: Hostname of the primary HANA instance to connect the system replication
         remote_instance:
           $name: Primary node instance number
           $type: text
           $optional: false
+          $help: Intance Number of the primary HANA instance to connect the system replication
```
arbulu89: A type here: Intance

form.yml
```diff
@@ -82,40 +90,44 @@ hana:
...
             $name: Backup file name
             $type: text
             $optional: false
+            $help: File name for the backup that will be created
         create_userkey:
-          $name: Create new User key
+          $name: Create new key (hdbuserstore)
           $type: boolean
           $default: false
+          $help: Mark this option to create a new key into the hdbuserstore.
```
arbulu89: I would remove this final dot as we don't have it in the other descriptions

form.yml
diegoakechi: Fixed.

PR #12, opened 2019-02-25
Choose shaptools py2 or py3 accordingly with the OS version.

diegoakechi:
SLE-15 has adopted python3 while SLE-12 still uses python2. So, the formula needs to install the correct python version shaptools package.


juadk:
Tested and it's works, but my python2 dependency problem seems to be elsewhere...

arbulu89:
@diegoakechi I'm not sure if this solves all the use cases. What about the cases where we need to install python2-shaptools (salt using python2 but SLES version is 15). Is this case possible?

Can we detect current python version during salt execution and use that to choose between versions?

diegoakechi:
@arbulu89 Nice work. Much better ;)

juadk:
@arbulu89 Great, this will fix my last salt issue!

PR #15, opened 2019-02-28
Hotfix/invalid name argument

arbulu89:
Fix the invalid name argument warning/error.
After salt fluorine version the name parameter must exist in salt state code

Besides that
- host value is used to set it in extra parameters dictionary to update properly the hostname in hana configuration file.
- keystore usage has been adapted to use key or user/password combos during the backup creation


hana/install.sls
```diff
@@ -6,7 +6,7 @@
 
 hana_install_{{ node.host+node.sid }}:
   hana.installed:
-    - sid: {{ node.sid }}
+    - name: {{ node.sid }}
```
diegoakechi: @arbulu89  is this change to name intentional? sid looks more appropriate to me...

arbulu89: @diegoakechi Yes, it is intentional. Due some changes in the last salt version (fluorine), name argument must be used as a first parameter. Otherwise an error is raise. 
I agree that **sid** would be more appropriate, but in order to make everything work in different salt version we are forced to update this (in fact, all of the methods of other salt states code have **name** as first parameter)

diegoakechi: @arbulu89 hum.. ok, sad, but I understand now... hehe.

PR #19, opened 2019-03-18
Hotfix/fix primary available dependency

arbulu89:
Cover the scenario where the secondary node is created before the primary and add a timeout to configure this timeout

### DIFF
--- a/hana/primary_available.sls
+++ b/hana/primary_available.sls
@@ -12,3 +12,6 @@ primary-available:
     - timeout: 100
     - require:
       - netcat-openbsd
+
+{% endif %}
+{% endfor %}
--- a/hana/primary_available.sls
+++ b/hana/primary_available.sls
@@ -9,7 +9,7 @@ netcat-openbsd:
 primary-available:
   cmd.run:
     - name: until nc -z {{  node.secondary.remote_host }} 40002; do sleep 1; done
-    - timeout: 100
+    - timeout: {{  node.secondary.primary_timeout|default(100) }}
     - require:
       - netcat-openbsd
 
--- a/hana/primary_available.sls
+++ b/hana/primary_available.sls
@@ -1,0 +1,14 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes %}
+{% if node.host == host and node.secondary is defined %}
+netcat-openbsd:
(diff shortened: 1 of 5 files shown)


hana/primary_available.sls
```diff
@@ -0,0 +1,17 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes %}
+{% if node.host == host and node.secondary is defined %}
+netcat-openbsd:
```
krig: actually I think I have a better way to do this in hacluster-formula where it looks for Hawk to be running, no need for netcat: https://github.com/SUSE/habootstrap-formula/blob/master/cluster/join.sls#L3

arbulu89: Agree! The other way looks better, less dependencies. I will make the update and test it.
Thanks @krig

krig: ah but couldn't you use the 7630 port here too?

krig: aha no you need to check if HANA is running of course, sorry!

arbulu89: Exactly, we need to check is System replication of HANA is enabled. There are many ports for that, but none of them uses http protocol

krig: I see yeah. Hm, seems salt doesn't come with anything better, at least at a glance..

hana/primary_available.sls
```diff
@@ -0,0 +1,17 @@
...
+  pkg.installed
+
+primary-available:
+  cmd.run:
+    - name: until nc -z {{  node.secondary.remote_host }} 40002; do sleep 1; done
```
diegoakechi: @arbulu89 The logic here is a little bit more complex. See: https://help.sap.com/viewer/ports/616a3c0b1cc748238de9c0341b15c63c.html

The port calculation depends if HANA is running on a Single Container or Multi-Container installation, also depends on the instance number like `4<InsNr>02`.

Maybe @Simranpal knows a command or a metadata file to check the port in usage without needing to calculate that.

arbulu89: True, I didn't notice this logic. Let's see if we can find a better solution to find the port.
Thx @diegoakechi

diegoakechi: As discussed, we will solve this with a different approach. Approving for now.

PR #20, opened 2019-03-22
[WIP] create srTakeover-Hook template

Simranpal:
parameters required by srTakeover-hook code: dbinst, dbport, dbuser, dbpassword

Things to improve:

1. Improve the node selection to fetch parameters, as currently retrieving all data from primary node as 'dbuser' exists in pillar data of primary node userkey

2. Currently putting the python code inside for loop as Jinja doesnt allow using loop variables outside the loop block, need to check if better solution exists


templates/srTakeover_hook.j2
```diff
@@ -0,0 +1,103 @@
...
+    {% set dbport = ('30'+ dbinst +'13')|int %}
+{% endif -%}
+
+#!/usr/bin/env python2
+
+"""
+Auto-generated HA/DR hook script
+
+"""
+
```
arbulu89: I would put here directly the parameters like:
```
dbuser="{{ dbuser }}"
dbpwd="{{ dbpwd }}"
...

We don't need to change anything else
```

templates/srTakeover_hook.j2
```diff
@@ -0,0 +1,103 @@
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes %}
+{% if node.host == host and node.primary is defined %}
```
arbulu89: Do we need to copy this file in the primary node?

Using this comparison you are only considering this template for the primary. If this must be used in the other machine (the machine we need to set as cost optimized) we will need to add an extra parameter to know where we have to create this file, and get the primary data too (maybe we need to loop twice, to get the cost optimized node and the primary)

Simranpal: Hi @arbulu89 This file need to be copied only to secondary node, and we need the user_name ('SYSTEM')which I could only find under the primary/userkey info, and was not under secondary info. 
Maybe in pillar data, we can add user_name on secondary, along with the extra parameter to distinguish for cost-optimized. What do you think?

templates/srTakeover_hook.j2
Simranpal: Hi @arbulu89  Sorry I did not understand, do you mean, that we don't need to compute these parameters from pillar file?

diegoakechi: What @arbulu89 mentioned here is that you don't need to interpolate the values in the entire script. That you can only initialize the variable on the beginning of the script, for example and use the normal python variable in all the occurrences... This way the code is cleaner and easy to change in case of need

arbulu89: Exactly, in fact the solution is super straight-forward. The script is ready to only change these 4 lines, nothing else (obviously we need to extract the data you do at the beginning)

templates/srTakeover_hook.j2
arbulu89: I think we don't need to add any extra user_name parameter. I would do something similar that we do in: https://github.com/SUSE/saphanabootstrap-formula/blob/master/hana/copy_ssfs.sls

First loop through the nodes to find the node where we have to create this file (looking for scenario-type: cost-optimized entry in pillar data for example). And then loop again to find the primary that matches with the remote host of the secondary. We can get all the data with that.

hana/defaults.yaml
```diff
@@ -41,3 +41,8 @@ hana:
         remote_instance: '00'
         replication_mode: 'sync'
         operation_mode: 'logreplay'
+        scenario_type: 'cost-optimized'
```
arbulu89: I wouldn't put these entries in the default.yml file. This is a quite specific deployment, so let's try to have it as generic as possible (,aybe we would need to remove some current values too)

hana/enable_cost_optimized.sls
```diff
@@ -0,0 +1,56 @@
...
+      - name: /hana/shared/srHook
+      - user: root
+      - mode: 755
+      - makedirs: True
+
+install_srTakeover_hook:
+    file.managed:
+      - source: salt://hana/templates/srTakeover_hook.j2
+      - name: /hana/shared/srHook/sr-Takeover.py
+      - user: root
```
arbulu89: Just a stupid question. Is the user root for these files or **sid**adm?

hana/enable_cost_optimized.sls
```diff
@@ -0,0 +1,56 @@
...
+      - group: root
+      - mode: 644
+      - template: jinja
+      - require:
+        - hana_install_{{ node.host+node.sid }}
```
arbulu89: It should depend on *reduce_memory_resources_{{  node.host+node.sid }}*, isn't it?

hana/enable_cost_optimized.sls
```diff
@@ -0,0 +1,56 @@
...
+        - setup_srHook_directory
+
+install_hana_python_packages:
+    archive.extracted:
+      - name: /hana/shared/srHook
+      - enforce_toplevel: False
+      - source: {{ grains['hana_inst_folder']+'/DATA_UNITS/HDB_CLIENT_LINUX_X86_64/client/PYDBAPI.TGZ' }}
```
arbulu89: We careful with this *+*, in jinja we must use *~* to concatenate strings

hana/install.sls
```diff
@@ -2,6 +2,11 @@
 {% set host = grains['host'] %}
 
 {% for node in hana.nodes %}
+{% if node.host == host and node.scenario_type is defined and node.scenario_type.lower() == 'cost-optimized' %}
+include_cost_optimized:
```
arbulu89: Is this working? I didn't know we could use *include* this way.

Anyway, maybe, we should consider just having an initial if with **node.host == host** for everything and then other if statements for *cost-optimized* and *install*. What do you think?

hana/enable_cost_optimized.sls
Simranpal: Hi @arbulu89 thanks for bringing this up. This is not documented in the guide.
As per my knowledge, the files under this directory should have updated permissions(740) and owner <sid>adm:sapsys 
I will update this accordingly.

hana/install.sls
Simranpal: Hi @arbulu89 I want to test the include this way using an id. The reason because I was getting error with my deployments last time " domain[1] found conflicting ID 'include' "

similar to salt issue https://github.com/saltstack/salt/issues/9400
There exists include already in 'enable_secondary.sls''

Below is relevant error:
-------------------------------------------------------------------------------------------

module.hana_node.module.hana_node.libvirt_domain.domain[0] (remote-exec): [INFO    ] Fetching file from saltenv 'base', ** done ** 'hana/enable_primary.sls'
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec): [INFO    ] Fetching file from saltenv 'base', ** done ** 'hana/install.sls'
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec): [CRITICAL] Rendering SLS 'base:hana.install' failed: while constructing a mapping
module.hana_node.module.hana_node.libvirt_domain.domain[0] (remote-exec): [INFO    ] Fetching file from saltenv 'base', ** done ** 'hana/enable_secondary.sls'
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec):   in "<unicode string>", line 8, column 1:
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec):     include:
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec):     ^
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec): found conflicting ID 'include'
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec):   in "<unicode string>", line 32, column 1:
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec):     include:
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec):     ^
module.hana_node.module.hana_node.libvirt_domain.domain[1] (remote-exec): [INFO    ] Fetching file from saltenv 'base', ** done ** 'hana/enable_primary.sls'

Simranpal: The issue was resolved after moving include out of loop

hana/enable_cost_optimized.sls
Simranpal: i set the correct ownership for the directory and files sidadm:sapsys

hana/enable_cost_optimized.sls
```diff
@@ -0,0 +1,58 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes %}
+{% if node.host == host and node.scenario_type is defined and node.scenario_type.lower() == 'cost-optimized' %}
+
+reduce_memory_resources_{{  node.host+node.sid }}:
+    hana.memory_resources_updated:
+      - name: {{  node.host }}
+      {% if node.cost_optimized_parameters is defined %}
```
arbulu89: If I'm not wrong the next 2 parameters are mandatory, right? Why do you check if this entry is defined. It must be defined. We should think about a better way to do that.
The formula even must fail if they are not defined, but the formula rendering must fail, not the `memory_resources_updated` call

hana/enable_cost_optimized.sls
```diff
@@ -0,0 +1,58 @@
...
+      - user_password: {{ node.install.system_user_password }}
+      {% endif %}
+      - sid: {{  node.sid }}
+      - inst: {{  node.instance }}
+      - password: {{  node.password }}
+      - require:
+        - hana_install_{{ node.host+node.sid }}
+{% endif %}
+
+{% if node.host == host and node.secondary is defined %}
```
arbulu89: How this if is working? If the node is set as secondary it will apply the changes??
What happens if the node is set as secondary but not as cost-optimized? Will this changes affect the node?

I guess, together of being secondary, cost-optimized flag must there too apply these changes

templates/srTakeover_hook.j2
```diff
@@ -0,0 +1,113 @@
...
+{% set dbinst = '{:0>2}'.format(node.instance) %}
+
+{% for prim_node in hana.nodes %}
+{% if node.secondary.remote_host == prim_node.host and prim_node.primary is defined %}
+
+{% set dbuser = prim_node.primary.userkey.user_name %}
+{% set dbpwd = prim_node.password %}
+{% set dbversion = salt['hana.get_version'](sid=dbsid, inst=dbinst, password=dbpwd) %}
+{% if salt['pkg.version_cmp'](dbversion,'2.0') < 0 %}
+    {% set dbport = ('3'+ dbinst +'15')|int %}
```
arbulu89: Strings are concatenated using `~` in jinja

templates/srTakeover_hook.j2
```diff
@@ -0,0 +1,113 @@
...
+            conn = dbapi.connect('localhost',dbport,dbuser,dbpwd)
+            cursor = conn.cursor()
+            cursor.execute(stmnt1)
+            cursor.execute(stmnt2)
+            return 0
+        elif rc == 2:
+            # error, something went wrong
+            return 0
+{% endif %}
+{% endfor %}
```
arbulu89: I think we can finish the prim `for` loop just after initializing the params, right? To avoid extra job

templates/srTakeover_hook.j2
```diff
@@ -0,0 +1,113 @@
...
+{% set dbsid = node.sid.lower() %}
+{% set dbinst = '{:0>2}'.format(node.instance) %}
+
+{% for prim_node in hana.nodes %}
+{% if node.secondary.remote_host == prim_node.host and prim_node.primary is defined %}
+
+{% set dbuser = prim_node.primary.userkey.user_name %}
+{% set dbpwd = prim_node.password %}
+{% set dbversion = salt['hana.get_version'](sid=dbsid, inst=dbinst, password=dbpwd) %}
+{% if salt['pkg.version_cmp'](dbversion,'2.0') < 0 %}
```
arbulu89: I would put an extra space after `dbversion,` to follow python rules

hana/enable_cost_optimized.sls
Simranpal: Hi Xabi, yes you are correct. 

here is summary of the logic I wanted:
- check if cost-optimized flag for this node pillar entry
   1) reduce resources on this hana sid
   2) check if this node is secondary - if so apply srHook stuff
-end if

I made a mistake there closing the previous endif early.
I originally intended to put this if condition within the previous condition where I check 'cost-optimized' flag.
The logic will be applied in that case. i need to close the initial condition correctly.
Let me know what you think of this approach, after I fix the above mistake.

templates/srTakeover_hook.j2
Simranpal: Hi @arbulu89 the variables are defined and used in the if condition under prim for loop. I will check if there is a better way to avoid extra work

hana/enable_cost_optimized.sls
arbulu89: Yes, this solution looks correct

templates/srTakeover_hook.j2
```diff
@@ -0,0 +1,112 @@
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes %}
+{% if node.host == host and node.secondary is defined and node.scenario_type.lower() == 'cost-optimized' %}
```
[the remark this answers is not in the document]
arbulu89: In fact, as this file will be rendered only in machines that are cost-optimized (this is already checked in enable_cost_optimized.sls (install_srTakeover_hook) we could only check `if node.host == host` and that's it, right?
What do you think?
We could avoid extra complexity in the if statement (but obviously, both works)

templates/srTakeover_hook.j2
arbulu89: What I only mean is that we don't need to have the whole script inside the for loops. We need the loops to get the variables data, that's it. That's why we could close the loops just after getting the needed data (after line 20 of the file)

hana/enable_cost_optimized.sls
```diff
@@ -0,0 +1,55 @@
...
+
+{% for node in hana.nodes %}
+{% if node.host == host and node.scenario_type is defined and node.scenario_type.lower() == 'cost-optimized' and node.cost_optimized_parameters is defined%}
+
+reduce_memory_resources_{{  node.host+node.sid }}:
+    hana.memory_resources_updated:
+      - name: {{  node.host }}
+      - global_allocation_limit: {{ node.cost_optimized_parameters.global_allocation_limit }}
+      - preload_column_tables: {{ node.cost_optimized_parameters.preload_column_tables }}
+      - user_name: SYSTEM
```
arbulu89: Is this SYSTEM entry correct hardcoded?

arbulu89:
Code merged in other PR

PR #21, opened 2019-04-08
Add support for GCP

juadk:
GCP isn't supporting [virtual IP yet:](https://issuetracker.google.com/issues/35904699)
By the way, I add a support for GCP in the performance_optimized scenario according to [google upstream script](https://storage.googleapis.com/sapdeploy/dm-templates/lib/sap_lib_ha.sh).

It works in my test but I'm not sure that's the optimized way to do it.

@diegoakechi @arbulu89 What do you think?

### DIFF
--- a/templates/performance_optimized.j2
+++ b/templates/performance_optimized.j2
@@ -17,19 +17,11 @@ op_defaults \
 # production HANA
 #
 
-{% if data.platform == "gcp" %}
 primitive rsc_ip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:IPaddr2 \
     params \
-    ip="{{ data.virtual_ip }}" cidr_netmask="32" nic="eth0" \
+    ip="{{ data.virtual_ip }}" cidr_netmask="{{ 32 if data.platform == "gcp" else data.virtual_ip_mask }}" {{ "nic=" ~ pillar.cluster.interface|default('eth0') }} \
     op start timeout="20" op stop timeout="20" \
     op monitor interval="10" timeout="20"
-{% else %}
-primitive rsc_ip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:IPaddr2 \
-    params \
-    ip="{{ data.virtual_ip }}" cidr_netmask="{{ data.virtual_ip_mask }}" \
-    op start timeout="20" op stop timeout="20" \
-    op monitor interval="10" timeout="20"
-{% endif %}
 
 primitive rsc_SAPHanaTopology_{{ sid }}_HDB{{ instance }} ocf:suse:SAPHanaTopology \
     params \
--- a/templates/performance_optimized.j2
+++ b/templates/performance_optimized.j2
@@ -17,11 +17,19 @@ op_defaults \
 # production HANA
 #
 
+{% if data.platform == "gcp" %}
+primitive rsc_ip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:IPaddr2 \
+    params \
+    ip="{{ data.virtual_ip }}" cidr_netmask="32" nic="eth0" \
+    op start timeout="20" op stop timeout="20" \
+    op monitor interval="10" timeout="20"
+{% else %}
 primitive rsc_ip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:IPaddr2 \
     params \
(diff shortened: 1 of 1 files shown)


templates/performance_optimized.j2
```diff
@@ -17,11 +17,19 @@ op_defaults \
...
+{% if data.platform == "gcp" %}
+primitive rsc_ip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:IPaddr2 \
+    params \
+    ip="{{ data.virtual_ip }}" cidr_netmask="32" nic="eth0" \
+    op start timeout="20" op stop timeout="20" \
+    op monitor interval="10" timeout="20"
+{% else %}
 primitive rsc_ip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:IPaddr2 \
     params \
     ip="{{ data.virtual_ip }}" cidr_netmask="{{ data.virtual_ip_mask }}" \
```
diegoakechi: @juadk if the only difference need is the cidr_netmask, I would consider an inline IF here to check the platform. This way, it would avoid repeating the entire block.  Another point is, for GCP is not needed to inform the nic into the resource? (AFAIK it assumes eth0 as default, but maybe to be more visually readable, I would include it here too).

arbulu89: Agree with @diegoakechi 

I would do something like that:
cidr_netmask="{{ 32 if data.platform == "gcp" else data.virtual_ip_mask }}"

About the nic, if this interface is the same than the used by crm, consider that libvirt might uses eth1

templates/performance_optimized.j2
```diff
@@ -17,19 +17,11 @@ op_defaults \
 # production HANA
 #
 
-{% if data.platform == "gcp" %}
 primitive rsc_ip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:IPaddr2 \
     params \
-    ip="{{ data.virtual_ip }}" cidr_netmask="32" nic="eth0" \
+    ip="{{ data.virtual_ip }}" cidr_netmask="{{ 32 if data.platform == "gcp" else data.virtual_ip_mask }}" {{ "nic=" + pillar.cluster.interface if data.platform == "gcp" }} \
```
diegoakechi: I think we could assume "eth0" as default, independent of the platform. I think is very common to have more than one interface, so for all platforms the customer can change the virtual IP to a specific interface.

```suggestion
    ip="{{ data.virtual_ip }}" cidr_netmask="{{ 32 if data.platform == "gcp" else data.virtual_ip_mask }}" {{ "nic=" ~ pillar.cluster.interface if pillar.cluster.interface else "eth0" }} \
```

P.S.: AFAIK, the `~` operator is the way to concatenate on Jinja.

juadk: @diegoakechi Thanks, let me try your suggested change. You are right for `~`.

arbulu89: @diegoakechi @juadk 
```
{{ "nic=" ~ pillar.cluster.interface or 'eth0' }}
```
might work as well (I didn't test it)

juadk: @diegoakechi @arbulu89 The both solutions work but if I remove `interface: eth0` from /srv/pillar/cluster.sls, templating fails with this error:
`Unable to manage file: Jinja variable 'salt.utils.odict.OrderedDict object' has no attribute 'interface'`

Maybe my syntax `pillar.cluster.interface` isn't the best way or we could consider that cluster.interface is mandatory for cluster deployment, AFAIK, it is.

diegoakechi: Well, it shouldn't be mandatory if we have a default =)

What if the inversion?

```suggestion
    ip="{{ data.virtual_ip }}" cidr_netmask="{{ 32 if data.platform == "gcp" else data.virtual_ip_mask }}" {{ "nic=" ~ "eth0" if not pillar.cluster.interface else pillar.cluster.interface }} \
```

arbulu89: @juadk 
Other solution would be:
```
{{ "nic=" ~ pillar.cluster.interface|default('eth0') }}
```

diegoakechi: This is much more elegant =)

juadk: And works better =)

juadk: Updated and tested with GCP as well as AWS.

PR #29, opened 2019-05-22
Update rsc template

Simranpal:
- Added the cluster resources for managing QAS HANA
- Renamed this template file to be more suitable for including cost-opt and performance opt scenarios


Simranpal:
> Hi @Simranpal ,
> 
> I don't really like how the non production HANA instance is managed. The assumption made to know where to apply the changes is quite fragile.
> 
> We need to think a better way to manage the cluster and hana pillar files. Now the, the data is imported from cluster pillar, but you added here the hana pillar too.
> Could we just use one file? You could add the needed entries to the cluster pillar file. I know, that they would be duplicated, but it is a way to keep the formula independent.
> 
> Actually, more I see this, I think that we should move the this template to the habootstrap-formula project. What do you think about that?

Hi @arbulu89 I made some changes to use only cluster pillar. 
And yes it makes sense to move this template to habootstrap-formula since it deals with cluster resources and most of the data is cluster pillar data.

PR #30, opened 2019-05-24
Feature/update secondary registration

arbulu89:


hana/defaults.yaml
```diff
@@ -41,3 +41,4 @@ hana:
         remote_instance: '00'
         replication_mode: 'sync'
         operation_mode: 'logreplay'
+        timeout: 100
```
diegoakechi: @arbulu89  we noticed in our tests that it can take much longer sometimes. Maybe something like 600 would be more appropriate as default?

PR #32, opened 2019-06-06
Hotfix/template

arbulu89:
Fix the HANA DB instance parameter usage


Simranpal:
Thanks for the fix @arbulu89 . We will need to request updating this in the SUSE guide/documentation as well

PR #38, opened 2019-06-13
[WIP] Fix for srTakeover_hook

Simranpal:
Updates to the srTakeover_hook
- Fixed the tarball extraction path for srHook related HANA py files
- Added the [ha_dr_provider_srTakeover] section to global.ini file 
for the hook to undo changes after takeover

Tested in local HANA environment, i need to test it with terraform deployment as well 
- Check that after takeover is triggered in cluster and when HANA on node02 become primary, the ini file changes(related to cost-optimized) are reverted/removed from global ini files:
/hana/shared/PRD/global/hdb/custom/config/global.ini

### DIFF
--- a/hana/enable_cost_optimized.sls
+++ b/hana/enable_cost_optimized.sls
@@ -32,7 +32,7 @@ setup_srHook_directory:
 install_srTakeover_hook:
     file.managed:
       - source: salt://hana/templates/srTakeover_hook.j2
-      - name: /hana/shared/srHook/sr-Takeover.py
+      - name: /hana/shared/srHook/srTakeover.py
       - user: {{ node.sid.lower() }}adm
       - group: sapsys
       - template: jinja
@@ -51,6 +51,20 @@ install_hana_python_packages:
       - require:
         - reduce_memory_resources_{{ node.host+node.sid }}
         - setup_srHook_directory
+
+configure_ha_dr_provider_srTakeover:
+    file.append:
+      - name:  /hana/shared/{{ node.sid }}/global/hdb/custom/config/global.ini
+      - text: |
+          [ha_dr_provider_srTakeover]
+          provider = srTakeover
+          path = /hana/shared/srHook
+          execution_order = 1
+      - require:
+        - reduce_memory_resources_{{ node.host+node.sid }}
+        - setup_srHook_directory
+        - install_srTakeover_hook
+        - install_hana_python_packages
 {% endif %}
 {% endif %}
 {% endfor %}
--- a/hana/enable_cost_optimized.sls
+++ b/hana/enable_cost_optimized.sls
@@ -54,7 +54,7 @@ install_hana_python_packages:
 
 configure_ha_dr_provider_srTakeover:
     file.append:
-      - name:  /hana/shared/{{ node.sid }}/global/hdb/custom/config/global.ini
+      - name:  /hana/shared/{{ node.sid.upper() }}/global/hdb/custom/config/global.ini
       - text: |
           [ha_dr_provider_srTakeover]
           provider = srTakeover
--- a/hana/enable_cost_optimized.sls
+++ b/hana/enable_cost_optimized.sls
@@ -47,6 +47,7 @@ install_hana_python_packages:
       - group: sapsys
       - enforce_toplevel: False
       - source: {{ grains['hana_inst_folder']~'/DATA_UNITS/HDB_CLIENT_LINUX_X86_64/client/PYDBAPI.TGZ' }}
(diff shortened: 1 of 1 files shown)


hana/enable_cost_optimized.sls
```diff
@@ -47,9 +47,24 @@ install_hana_python_packages:
       - group: sapsys
       - enforce_toplevel: False
       - source: {{ grains['hana_inst_folder']~'/DATA_UNITS/HDB_CLIENT_LINUX_X86_64/client/PYDBAPI.TGZ' }}
```
arbulu89: Hi @Simranpal ,
I have found one thing here that we should change. Using this `grains` we totally depend in the terraform deployment, and we cannot do that.
We need to find other way of passing this `hana_inst_folder` data. It might be in the pillar file (i don't know if this could be autodetected somehow).
Let's find the best solution for that

diegoakechi: @Simranpal @arbulu89 Does this path work? The hana_inst_folder contains the NFS share location, so it would need to be mounted. This path should be composed using the pillar variable `software_path:` that contains the already mounted path of the SAP binaries.
Another important point is that the path only works for x86_64 platform. It should be composed based on the architecture being used, to work on Power too.

arbulu89: @diegoakechi Here we assume that the files are already in that path. We have to consider that they might be downloaded because we are deploying in some cloud provider.
Other feature would be just to get the installation files in this formula.

Actually, we have had hardcoded even in shaptools this path. We need to make it variable even there (right now we are only using x86)

diegoakechi: > @diegoakechi Here we assume that the files are already in that path. We have to consider that they might be downloaded because we are deploying in some cloud provider.
The client is provided as part of the HANA Installer as a whole. So it should use the same directory that was mounted in order to perform the HANA installation, that is the value included on the `software_path:` entry on the `hana.sls` pillar file. 

> Other feature would be just to get the installation files in this formula.
> 
> Actually, we have had hardcoded even in shaptools this path. We need to make it variable even there (right now we are only using x86)
Yes, this is a problem but should be easily solved with something like `os.uname()` or `platform.machine()`.

Simranpal: Hi @arbulu89 @diegoakechi I will use the pillar data entry("software_path") for fetching the client py files.

For the platform specific HANA client path, we need to store the platform-type first in grains? and then maybe based on that update the salt logic..

diegoakechi: @Simranpal Salt already gets it on the os core grains. I think the grains['cpuarch'] could be used for that. The problem is, that if we need to compose it in multiple places, maybe would make more sense to move it to shaptools as a method, or as a function inside the salt modules. What do you think @arbulu89?

arbulu89: @diegoakechi @Simranpal Actually we will need to do this updates in shaptools too. By now, forget about the CPU arch, we will manage this in a next PR.
We need to be careful to with this, maybe managing this in the formula is not a good idea (neither in shaptools) we will see

hana/enable_cost_optimized.sls
```diff
@@ -46,10 +46,25 @@ install_hana_python_packages:
...
+      - source: {{ node.install.software_path~'/DATA_UNITS/HDB_CLIENT_LINUX_X86_64/client/PYDBAPI.TGZ' }}
+      - options: --strip=1 --wildcards 'hdbcli/*.py'
       - require:
         - reduce_memory_resources_{{ node.host+node.sid }}
         - setup_srHook_directory
+
+configure_ha_dr_provider_srTakeover:
+    file.append:
+      - name:  /hana/shared/{{ node.sid.upper() }}/global/hdb/custom/config/global.ini
+      - text: |
```
arbulu89: I would add a new line before the text, otherwise th output is quite ugly

```
mode = sync
actual_mode = sync
site_name = SECONDARY_SITE_NAME
operation_mode = logreplay
[ha_dr_provider_srTakeover]
provider = srTakeover
path = /hana/shared/srHook
execution_order = 1
```
Not really visual

arbulu89:
And other thing. I think the srTakeover.py is not correct. The initial parameters are written before the shebang. We have to put them after that (`#!/usr/bin/env python2`)

Simranpal:
This PR can removed as these changes are included in newer PR: https://github.com/SUSE/saphanabootstrap-formula/pull/49

arbulu89:
Closing PR. Same functionality implemented in: https://github.com/SUSE/saphanabootstrap-formula/pull/49

PR #42, opened 2019-07-03
Execute hanadb_exporter as a daemon

arbulu89:
Execute hanadb_exporter as a daemon.
And hanadb_exporter installation is suggested (even though it is installed in the salt execution itself)

### DIFF
--- a/hana/exporter.sls
+++ b/hana/exporter.sls
@@ -5,7 +5,7 @@
 {% if node.host == host and node.exporter is defined %}
 
 {% set instance = '{:0>2}'.format(node.instance) %}
-{% set config_file = '/etc/hanadb_exporter/config_{}_{}.json'.format(node.sid, instance) %}
+{% set config_file = '/etc/hanadb_exporter/{}_{}.json'.format(node.sid, instance) %}
 
 hanadb_exporter:
   pkg.installed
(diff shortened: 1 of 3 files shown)


hana/exporter.sls
```diff
@@ -5,7 +5,8 @@
 {% if node.host == host and node.exporter is defined %}
 
 {% set instance = '{:0>2}'.format(node.instance) %}
-{% set config_file = '/etc/hanadb_exporter/config_{}_{}.json'.format(node.sid, instance) %}
+{% set daemon_instance = '{}_{}'.format(node.sid, instance) %}
```
MalloZup: I personally prefer `{0} {1}` it make the code imho more implicit. But is just a personal style, ok with both.

hana/exporter.sls
```diff
@@ -22,15 +23,11 @@ configure_exporter:
...
-stop_exporter:
-  process.absent:
-    - name: hanadb_exporter -c {{ config_file }} -m /etc/hanadb_exporter/metrics.json
-    - require:
-        - configure_exporter
-
 start_exporter:
-  cmd.run:
-    - name: nohup hanadb_exporter -c {{ config_file }} -m /etc/hanadb_exporter/metrics.json &>/dev/null &
+  service.running:
```
MalloZup: in theory we could also `retry` the service start .. but imho we can just avoid for moment  this :grin:

hana/exporter.sls
arbulu89: @MalloZup Do you mean `{0}_{1}`? As this extension will be used to match with a file name, we cannot have spaces.

MalloZup: @arbulu89 I mean the format thing in python. you can write `{0}` or `{}` they are identic. 

but is same.. :grin:

MalloZup: I'm resolving now

hana/exporter.sls
MalloZup: resolving

PR #43, opened 2019-07-05
Retry-pkg installation if it fails 

MalloZup:
# desc:
- add retry times.
- updated spec and changelog.

https://trello.com/c/rp9qZp6B/761-retry-package-installation-if-it-fails

# Note for reviewers:
- I might need to test this with vagrant and other utils in case, should be ok, just in case I didn't overlooked some yaml spaces.

### DIFF
--- a/hana/exporter.sls
+++ b/hana/exporter.sls
@@ -9,10 +9,16 @@
 {% set config_file = '/etc/hanadb_exporter/{}.json'.format(daemon_instance) %}
 
 hanadb_exporter:
-  pkg.installed
+  pkg.installed:
+  - retry:
+      attempts: 3
+      interval: 15
 
 python3-PyHDB:
-  pkg.installed
+  pkg.installed:
+  - retry:
+      attempts: 3
+      interval: 15
 
 configure_exporter:
   file.managed:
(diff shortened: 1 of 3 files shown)


hana/exporter.sls
```diff
@@ -9,9 +9,15 @@
 
 hanadb_exporter:
   pkg.installed
```
arbulu89: I think colons are missing here

MalloZup: the colons are needed if you add `name` later, but here since hanadb_exporter is the argument of pkg.installed we don't need them.. 

https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkg.html

MalloZup: or maybe, i might need to setup vagrant for that..

hana/exporter.sls
```diff
@@ -9,10 +9,16 @@
...
-  pkg.installed
+  pkg.installed:
+  - retry:
+      attempts: 3
+      interval: 15
 
 python3-PyHDB:
-  pkg.installed
+  pkg.installed:
+ - retry:
```
arbulu89: Could you indent properly this `retry` entry

MalloZup: yep thx.. doing.. :crying_cat_face:

PR #45, opened 2019-07-08
Feature/power support

arbulu89:
Add support for Power machines

https://trello.com/c/AlM5SCiB/765-add-support-for-power-machines


MalloZup:
@arbulu89  we could perhaps merge this one before? https://github.com/SUSE/saphanabootstrap-formula/pull/43

arbulu89:
Tested and working for intel machines

hana/get_platform.sls
```diff
@@ -0,0 +1,18 @@
+#!py
+import platform
+
+PLATFORMS = {
+    'x86_64': 'HDB_CLIENT_LINUX_X86_64',
+    'ppc64': 'HDB_CLIENT_LINUX_PPC64',
+    'ppc64le': 'HDB_CLIENT_LINUX_PPC64LE'
+}
```
Simranpal: Hi @arbulu89 you added the HDB_CLIENT_ prefix to PLATFORMS.
It works for our scenario but maybe this is not expected result if just fetching the platform type

arbulu89: Hi @Simranpal I don't really understand your point.
What should we put there in your opinion?
As far as I have seen in all of our SAP installation folder, these are the names used by SAP for these platforms

Simranpal: Hi @arbulu89 what I mean is get_platform.sls should return the platform type x86_64/ ppc64 etc. Currently, this returns the directory name for HDB Client (thats what is required now).

I meant to correct the names, I think here filename or the PLATFORMS variable could be renamed to something else, like HDB_CLIENT TYPE

MalloZup: I think  that `PLATFORMS` is ok, imho.

If we want to make more precision: `HDB_CLIENT_PLATFORMS`  imho

`HDB_CLIENT_TYPE` type is to generic word, and one could understand as type of client (http/whatever protocols).

arbulu89: So, the discussion is about the naming, right?
Ok, I would think other way of renaming the sls file and the code.
Besides that, what do you think about the functionality @Simranpal ?

hana/get_platform.sls
```diff
@@ -0,0 +1,18 @@
+#!py
+import platform
+
+PLATFORMS = {
+    'x86_64': 'HDB_CLIENT_LINUX_X86_64',
+    'ppc64': 'HDB_CLIENT_LINUX_PPC64',
```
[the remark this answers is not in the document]
arbulu89: Thanks for your comment! 
I saw that SAP provides the installer for ppc64, so that's why I added it here.
We can remove it @diegoakechi what do you think?

hana/get_platform.sls
Simranpal: The functionality looks good @arbulu89

hana/get_platform.sls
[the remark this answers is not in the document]
diegoakechi: @arbulu89 Yes, it is not supported by SUSE and AFAIK it is not supported by SAP anymore. It is there for backward compatibility. We can remove it.

hana/get_platform.sls
diegoakechi: @arbulu89 I think we are mixing platform with HANA components here IMO.  The directory name determines the HANA component that it has inside + the platform where it was built. as HANA components we have `HANA Client`, `HANA Server`, `HANA XServer` and many others, but currently we install only some of these components. So, in that sense, I would split the directory component name, from the arch that composes the name, so we can reuse it in the future for the new components that we eventually will install. (they will come for S/4 HANA for exmple).

arbulu89: Ok. So I will change the code to only return the platform (formatted to capital letters) and create the path to the client with that.

arbulu89:
I have found an issue during the usage of the new way of getting the platform...
An error is raised because shaptools is not installed when the rendering is done, so the `get_platform.sls` rendering fails... Looking for a fix:

```
ModuleNotFoundError: No module named 'shaptools'

```

arbulu89:
Some changes done. Please, have a look on them!

PR #47, opened 2019-07-22
Fix issue with file permissions for SUMA

arbulu89:
Tested in environment with SUMA and the formulas are installed properly without breaking SUMA webpage.

INFO: uninstalling the package still removes `/usr/share/salt-formulas` folder breaking SUMA webpage


saphanabootstrap-formula.spec
```diff
@@ -103,9 +105,9 @@ fi
 %{fdir}/states/%{fname}/%{ftemplates}
 %{fdir}/metadata/%{fname}
 
-%dir %attr(0750, root, salt) %{fdir}
```
diegoakechi: @arbulu89 I think it will conflict with the `salt-standalone-formula-configuration` package as it creates the directory with different permission. We need to set specific attribute values for the files that will be different from the directories permissions and keep these directory permissions. AFAIK chaging the default attribute to `%defattr(-,root,root)` should do the trick...

saphanabootstrap-formula.spec
```diff
@@ -32,21 +35,20 @@ Requires:       habootstrap-formula
 Requires:       salt-shaptools
 Suggests:       hanadb_exporter >= 0.3.1
 
-# On SLE/Leap 15-SP1 and TW requires the new salt-formula configuration location.
-%if ! (0%{?sle_version:1} && 0%{?sle_version} < 150100)
```
diegoakechi: This cannot be dropped, as this package already exists on TW, so the automatic formula configuration is already possible there. Also, having this dependency helps to avoid file conflicts and dependency problems between the packages. It was only removed on the PackageHub submission as this package is not yet available there. As soon as it gets included there, the same TW package will be synced on PackageHub.

arbulu89: So, in this case, we will manage the packageHub spec file manually, right?

diegoakechi: Yes, only for now.

MalloZup:
do we have a bug for suma against this failure?

arbulu89:
> do we have a bug for suma against this failure?

Yes, 2 tickets were created:
https://bugzilla.suse.com/show_bug.cgi?id=1142309
https://bugzilla.opensuse.org/show_bug.cgi?id=1142306

PR #49, opened 2019-08-06
Fix/sr hook for cost-optimized

Simranpal:
Contains the fixes for the sr hook to undo changes cost-optimized related parameters on secondary HANA node after takeover

- Make the hdbcli client py files extraction compatible with new HANA revisions, as tarball has different structure in older HANA revisions
- Added the [ha_dr_provider_srTakeover] section to global.ini file 
- Other minor fixes to naming
- Added improvements suggested from other PR https://github.com/SUSE/saphanabootstrap-formula/pull/38

Tested in local environment

### DIFF
--- a/hana/enable_cost_optimized.sls
+++ b/hana/enable_cost_optimized.sls
@@ -28,6 +28,8 @@ setup_srHook_directory:
       - group: sapsys
       - mode: 755
       - makedirs: True
+      - require:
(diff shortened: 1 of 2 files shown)


Simranpal:
Structure of PYDBAPI.TGZ and hdbcli tar.gz in new HANA revision
image

Simranpal:
Hi @arbulu89 Thanks. I added the state to remove the tar file and hdbcli directory

PR #52, opened 2019-08-27
simplify installation of pattern

### LINKED ISSUE #51: simplify pkg installation especiall pattern
https://github.com/SUSE/saphanabootstrap-formula/blob/master/hana/packages.sls#L9

## error
`
module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): [CRITICAL] Rendering SLS 'base:cluster.packages' failed: Problem running salt function in Jinja template: Zypper command failure: Check Zypper's logs.; line 11   module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): --- module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): [...] module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): {% if grains['os_family'] == 'Suse' %} module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): {% set pattern_available = salt['cmd.retcode']('zypper search patterns-ha-ha_sles') %} module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): {% endif %}   module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): {% if pattern_available == 0 %} module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): {% set repo = salt['pkg.info_available']('patterns-ha-ha_sles')['patterns-ha-ha_sles']['repository'] %}    <====================== module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): patterns-ha-ha_sles: module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec):   pkg.installed: module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec):     - fromrepo: {{ repo }} module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec):     - retry: module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec):         attempts: 3 module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): [...] module.hana_node.null_resource.hana_node_provisioner[1] (remote-exec): --- `
`
# proposed solution
we should simplify this by just installing it without `jinja` functions.
(to do : https://github.com/uyuni-project/sumaform/blob/e9664647208498cf5b0b79695fa647c37a172dcf/salt/suse_manager_server/init.sls#L32)

this will fix also issue https://github.com/SUSE/ha-sap-terraform-deployments/issues/140

MalloZup:
Remove the jinja construct and simplify. The goal is to provide better
error because when it fails, it doesn't show any useful error before
with the jinja construct.

I need to test this

fix #51

# TODO (another pr)

Backport this to the other formula here https://github.com/SUSE/habootstrap-formula/blob/9ac90aaeed3d85276960c49db6d4fd9293c67ff3/cluster/packages.sls#L11

### DIFF
--- a/saphanabootstrap-formula.spec
+++ b/saphanabootstrap-formula.spec
@@ -19,7 +19,7 @@
 # See also http://en.opensuse.org/openSUSE:Specfile_guidelines
 
 Name:           saphanabootstrap-formula
-Version:        0.2.9
+Version:        0.3.0
 Release:        0
 Summary:        SAP HANA platform deployment formula
 License:        Apache-2.0
--- a/hana/packages.sls
+++ b/hana/packages.sls
@@ -1,17 +1,12 @@
 #required packages to install SAP HANA
 
-{% set pattern_available = 1 %}
 {% if grains['os_family'] == 'Suse' %}
-{% set pattern_available = salt['cmd.retcode']('zypper search patterns-sap-hana') %}
-{% endif %}
-
-{% if pattern_available == 0 %}
-{% set repo = salt['pkg.info_available']('patterns-sap-hana')['patterns-sap-hana']['repository'] %}
-patterns-sap-hana:
-  pkg.installed:
-    - fromrepo: {{ repo }}
+install-patterns-sap-hana:
+  pkg.latest:
+    - name: patterns-sap-hana
+    - refresh: True
     - retry:
-        attempts: 3
+        attempts: 5
         interval: 15
 
 {% else %}
--- a/saphanabootstrap-formula.changes
+++ b/saphanabootstrap-formula.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug 28 10:03:32 UTC 2019 - Dario Maiocchi <dmaiocchi@suse.com>
+
+- Simplify pattern installation of formula
+  * remove redundant checks, that shadow erorr in jinja template
+  * improve debug UX 
+
+-------------------------------------------------------------------
 Wed Aug  7 07:46:25 UTC 2019 - Simranpal Singh <simranpal.singh@suse.com>
 
 - Version bump 0.2.9


MalloZup:
@ayoub-belarbi feel free to try this out also

ayoub-belarbi:
@arbulu89 Can you review this ? I don't know why that code existed in the first place

saphanabootstrap-formula.spec
```diff
@@ -19,7 +19,7 @@
 # See also http://en.opensuse.org/openSUSE:Specfile_guidelines
 
 Name:           saphanabootstrap-formula
-Version:        0.2.9
+Version:        0.3.0
```
arbulu89: Additionally, I wouldn't increase the version this way. I would update to `0.2.10` as this is just a bugfix or really minor update.

MalloZup: yep

MalloZup:
> We need to be careful with this change. This is the reason why the previous code was done:
> 
> The installation of the pattern will only work if the system is registered to SCC, and to the HA extension. That's why we perform a `search` initially. To check if the pattern is available. As currently the registration is not mandatory (even though we always do) the pattern won't be found in all cases.
> That's why we need a way to know if the pattern is available or not.
> With the new change, if the system is not registered, it will always fail (as the pattern is not available).
> 
> For that reason I think we cannot merge this changes. If you have any better idea about how to deal with this reason we can obviously discuss.
> 
> One option could be to use the `onlyIf` or `unless` options there, and depending of the return of the search, perform the pattern installation or the single packages installation.
> 
> Some comments may help as well to understand the reason behind this implementation, but to add comments in jinja is not so user friendly hehe

I think we should assume as `basis` that the system is always regiestered for the formulas. This assumption is imho

MalloZup:
the current question is here, why we assume that `the registration is not mandatory` ? to me should be a requirement, e.g we should assume that the repositories for the pkg are in place.

Or I'm missing something? :thinking:

MalloZup:
I'm hitting this error again

[ERROR   ] retcode: 105
[ERROR   ] An exception occurred in this state: Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/salt/state.py", line 1939, in call
    ret = self.states[cdata['full']](*cdata['args'], **cdata['kwargs'])
  File "/usr/lib/python3.6/site-packages/salt/loader.py", line 1941, in wrapper
    return f(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/salt/states/pkg.py", line 1539, in installed
    pkgs, refresh = _resolve_capabilities(pkgs, refresh=refresh, **kwargs)
  File "/usr/lib/python3.6/site-packages/salt/states/pkg.py", line 946, in _resolve_capabilities
    ret = __salt__['pkg.resolve_capabilities'](pkgs, refresh=refresh, **kwargs)
  File "/usr/lib/python3.6/site-packages/salt/modules/zypperpkg.py", line 2503, in resolve_capabilities
    refresh_db()
  File "/usr/lib/python3.6/site-packages/salt/modules/zypperpkg.py", line 1188, in refresh_db
    out = __zypper__.refreshable.call('refresh', '--force')
  File "/usr/lib/python3.6/site-packages/salt/modules/zypperpkg.py", line 340, in __call
    raise CommandExecutionError('Zypper command failure: {0}'.format(self.error_msg))
salt.exceptions.CommandExecutionError: Zypper command failure: Check Zypper's logs.
```
I think that one of big problem of having the refresh during the `jinja` is that we don't fail. If for example for some reason we  cannot find/install the pattern, we will just continue the installation which shouldn't be

**Comment by arbulu89** [comment]
Hi @MalloZup,
I have been researching about easy solutions and the next would be interesting:
```
{% set repo = salt['pkg.info_available']('patterns-sap-nw', refresh=False)['patterns-sap-nw']['repository'] %}
```
We could just add the `refresh=False` option to the `pkg.info_available` call. Right now, when it fails, the failure comes during the refresh operation.
I will test this implementation, and if it works we could see if it's a good final option. In that case we should expand it to the other salt formulas

MalloZup:
closing

PR #54, opened 2019-10-25
Add support for multi tenant hanadb_exporter

arbulu89:
Add support for multi tenant hanadb_exporter.

Just add the new values to the daemon configuration file and SUMA form.


form.yml
```diff
@@ -232,7 +232,11 @@ hana:
         exposition_port:
           $name: SAP HANA exposter exposition port
           $type: text
-          $optional: true
+          $optional: false
+        multi_tenant:
+          $name: Enable Multi tenant monitoring
+          $type: boolean
```
diegoakechi: @arbulu89 if possible, I would include a `$help` here explaining that, if enabled, the exporter will select all the tenants listed on the SYSTEMDB metadata and open a connection to each of them in order to collect metrics...

form.yml
```diff
@@ -242,3 +246,14 @@ hana:
           $name: SAP HANA password
           $type: password
           $optional: false
+        port:
+          $name: HANA database port. Only used when multi-tenant value is False
```
diegoakechi: @arbulu89 This can be a problem. By default the SYSTEMDB runs on 3<instNr>13 but the user can point it to any other different port, so even on the multi-tenant option, the user should be able to inform in what port the SYSTEMDB is running and then the exporter will automatically discover the ports of the other Tenant DBs.

arbulu89: Ok, I will set the port option customizable for the multi-tenant option

PR #57, opened 2019-11-27
Change Cluster Configuration Template to use `socat` inplace of `netcat` for Azure

Simranpal:
Azure changed the Load-Balancer listener to socat as netcat does not allow more than one concurrent connection. For that, we need to install the package socat :
https://github.com/SUSE/habootstrap-formula/pull/33
and correspondingly, our template for HANA (and Netweaver in the future) needs to be adapted to use the updated command, instead of the old nc command.

```
crm configure primitive rsc_nc_HN1_HDB03 anything \
params binfile="/usr/bin/socat" cmdline_options="-U TCP-LISTEN:62503,backlog=10,fork,reuseaddr /dev/null" \
op monitor timeout=20s interval=10 depth=0
```
Reference:
https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources

### DIFF
--- a/templates/scale_up_resources.j2
+++ b/templates/scale_up_resources.j2
@@ -53,8 +53,7 @@ ms msl_SAPHana_{{ sid }}_HDB{{ instance }} rsc_SAPHana_{{ sid }}_HDB{{ instance 
 {% if data.platform == "azure" %}
 
 primitive rsc_nc_{{ sid }}_HDB{{ instance }} anything \
-  params binfile="/usr/bin/nc" cmdline_options="-l -k 62503" \
-  meta resource-stickiness=0 \
+  params binfile="/usr/bin/socat" cmdline_options="-U TCP-LISTEN:62503,backlog=10,fork,reuseaddr /dev/null" \
   op monitor timeout="20" interval="10" depth="0"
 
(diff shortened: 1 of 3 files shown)


templates/scale_up_resources.j2
```diff
@@ -53,8 +53,7 @@ ms msl_SAPHana_{{ sid }}_HDB{{ instance }} rsc_SAPHana_{{ sid }}_HDB{{ instance
 {% if data.platform == "azure" %}
 
 primitive rsc_nc_{{ sid }}_HDB{{ instance }} anything \
-  params binfile="/usr/bin/nc" cmdline_options="-l -k 62503" \
-  meta resource-stickiness=0 \
+  params binfile="/usr/bin/socat" cmdline_options="-U TCP-LISTEN:62503,backlog=10,fork,reuseaddr /dev/null" \
```
arbulu89: @Simranpal I think the `TCP-LISTEN` port cannot be hardcoded as the last 2 digits are the HANA instance number.
Could you confirm that?
It should be `TCP-LISTEN:625{{ instance }}`

Simranpal: Hi @arbulu89 yes this seems to be the HANA probe port which is based on the instance number.

https://documentation.suse.com/sbp/all/pdf/SBP-SAP-AzureSolutionTemplates_color_en.pdf
here where hana instance is 00

arbulu89:
@Simranpal Before merging, please, upgrade package version and changes file!!!

PR #58, opened 2019-12-10
Remove default value to nic in scale up template

arbulu89:
Fix `nic` usage.
The logic is.
If not gcp, use interface if exists, otherwise don't set nic
If gcp, use interface if exists, otherwise default is `eth0`

PD: I have rename `nc` to `socat` in the template as makes more sense now


templates/scale_up_resources.j2
```diff
@@ -52,11 +54,11 @@ ms msl_SAPHana_{{ sid }}_HDB{{ instance }} rsc_SAPHana_{{ sid }}_HDB{{ instance
 
 {% if data.platform == "azure" %}
 
-primitive rsc_nc_{{ sid }}_HDB{{ instance }} anything \
+primitive rsc_socat_{{ sid }}_HDB{{ instance }} anything \
```
Simranpal: Can we rename this resource as such? Just thinking where else the resource name is used

arbulu89: I have changed the name in all the dependencies.
We should do the same in the sapnwbootstrap-formula

PR #60, opened 2019-12-10
Add new options to get primary node password

arbulu89:
The formula has an issue. If the primary node is not defined the secondary cannot be provisioned as it depends in the primary node password. This fixes this issue adding new options to get the primary password.
The options:
The primary password is retrieved in this order
1. If the primary node is defined in the pillar, primary password will be used
2. If secondary.primary_pass is defined this password will be used
3. The secondary machine password will be used

### DIFF
--- a/hana/enable_secondary.sls
+++ b/hana/enable_secondary.sls
@@ -3,9 +3,21 @@
 
 {% for node in hana.nodes if node.host == host and node.secondary is defined %}
 
-{% for prim_node in hana.nodes %}
-{% if node.secondary.remote_host == prim_node.host and prim_node.primary is defined %}
-{% set primary_pass = prim_node.password %}
+# The primary password is retrieved in this order
+# 1. If the primary node is defined in the pillar, primary password will be used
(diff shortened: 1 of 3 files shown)


hana/enable_secondary.sls
```diff
@@ -18,8 +30,7 @@
     - operation_mode: {{ node.secondary.operation_mode }}
     - timeout: {{ node.secondary.primary_timeout|default(100) }}
     - interval: {{ node.secondary.interval|default(10) }}
-    - primary_pass: {{ primary_pass }}
+    - primary_pass: {{ password.primary }}
```
Simranpal: Hi @arbulu89 does the password variable need to be defined as dict here, I was wondering if it works as such

arbulu89: Yes, this is the way. Actually, I was forced to use a dictionary, otherwise the variables defined in a for loop only have local scope, so you cannot use them outside. That's why the dictionaries are used here.

PR #64, opened 2019-12-20
Pass xml file to install hana if this file is provided

Simranpal:
Add the option to pass xml file containing passwords for installing hana.
system_user_password/sapadm_password in pillar will used if XML file is not provided
Related PR:
https://github.com/SUSE/shaptools/pull/48
https://github.com/SUSE/salt-shaptools/pull/46

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -13,7 +13,7 @@ hana:
         root_user: 'root'
         root_password: 's'
         # Fetch HANA passwords from XML file 
-        hdb_pwd_file: '/root/passwords.xml'
+        hdb_pwd_file: 'salt://passwords.xml'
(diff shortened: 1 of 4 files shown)


pillar.example
```diff
@@ -12,6 +12,9 @@ hana:
         software_path: '/root/sap_inst/51052481'
         root_user: 'root'
         root_password: 's'
+        # Fetch HANA passwords from XML file 
+        hdb_pwd_file: '/root/passwords.xml'
```
arbulu89: As the password must be obtained from the salt master file directory, I would change the example to something like:
`hdb_pwd_file: 'salt://passwords.xml'`

PR #65, opened 2020-01-21
Add saptune solution apply to formula

MalloZup:
implement latest module enablement to the formula via pillar

### DIFF
--- a/hana/saptune.sls
+++ b/hana/saptune.sls
@@ -1,0 +1,14 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes if node.host == host and (hana.saptune_solution is defined or node.saptune_solution is defined) %}
+
+{% set saptune_solution = node.saptune_solution|default(hana.saptune_solution) %}
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set name = '{}_{}'.format(node.sid, instance) %}
+
+apply_saptune_solution_{{ host }}_{{ name }}:
+  saptune.solution_applied:
+    - name: {{ saptune_solution }}
+
+{% endfor %}
--- a/pillar.example
+++ b/pillar.example
@@ -3,6 +3,12 @@ hana:
   # If set to false, this packages must be installed before the formula
   # execution manually
   # install_packages: true
+
+  # saptune solution to apply to all nodes ( by default nothing is applied)
+  # you can also use this to a single node if need to differ. see hana2 
+  # Warning: only a unique solution can exist into a node.
+  saptune_solution: 'HANA'
+
   nodes:
     - host: 'hana01'
       sid: 'prd'
@@ -46,6 +52,7 @@ hana:
       sid: 'prd'
       instance: 00
       password: 'Qwerty1234'
+      saptune_solution: 'MAXDB'
       install:
         software_path: '/root/sap_inst/51052481'
         root_user: 'root'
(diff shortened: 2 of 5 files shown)


MalloZup:
I will test this right now but imho it could work

pillar.example
```diff
@@ -8,6 +8,8 @@ hana:
       sid: 'prd'
       instance: 00
       password: 'Qwerty1234'
+      # saptune solution to apply ( by default nothing is applied)
+      saptune_solution: 'MAXDB'
```
arbulu89: Set `HANA` in the example. It makes more sense as this formula is about HANA

saphanabootstrap-formula.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Jan 21 14:41:29 UTC 2020 - Dario Maiocchi <dmaiocchi@suse.com>
+
+-  Version 0.4.3
+  * Add saptune module for apply a specified by user solution to each node
```
arbulu89: Better `Add saptune module to apply a specific solution`. What do you think?

MalloZup:
@arbulu89 regarding the suggested change, imho looked simply my Implementation (each node has independeance parameter) rather then having a global mechanism for all nodes and overwritten mechanism if a node want to be different. 

But i'm ok to do also the other way  thx for review !

pillar.example
MalloZup: done thc

MalloZup:
@arbulu89 all changes applied as requested.

I have tested the changes manually and was fine. :robot:

pillar.example
```diff
@@ -72,7 +76,7 @@ hana:
         user: 'SYSTEM'
         password: 'Qwerty1234'
 
-    - host: hana02
+    - host: hana03
```
arbulu89: Don't change this. This is to show that you can have more than 1 SAP instance in the same host (this is the cost optimized scenario in fact)

hana/saptune.sls
```diff
@@ -0,0 +1,12 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes if node.host == host and hana.saptune_solution is defined %}
```
arbulu89: I don't know if I would check `hana.saptune_solution is defined`.
And if it's not, but we have the solution in the node. This wouldn't work.
In any case we would need: 
```
{% for node in hana.nodes if node.host == host and (hana.saptune_solution is defined or node.saptune_solution is defined) %}
```

hana/saptune.sls
```diff
@@ -0,0 +1,11 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes if node.host == host and (hana.saptune_solution is defined or node.saptune_solution is defined) %}
+{% set saptune_solution = node.saptune_solution|default(hana.saptune_solution) %}
+
+apply_saptune_solution_{{node.host}}{{loop.index}}:
```
arbulu89: I think is better to have something like:
```
{% set instance = '{:0>2}'.format(node.instance) %}
{% set name = '{}_{}'.format(node.sid, instance) %}
apply_saptune_solution_{{ host }}_{{ name }}:
```

pillar.example
```diff
@@ -3,6 +3,10 @@ hana:
   # If set to false, this packages must be installed before the formula
   # execution manually
   # install_packages: true
+
+  # saptune solution to apply to all nodes ( by default nothing is applied)
```
arbulu89: Add the example of how to use `saptune_solution` in a node entry. And besides that, write a warning message that only a unique saptune solution can exist in a node, and if they are different, the formula will fail

MalloZup: ok done

PR #66, opened 2020-01-30
Feature/aws resources

arbulu89:
AWS resource and fence agents added; Here the guide (chapter 6.3):
https://www.suse.com/media/guide/suse_linux_enterprise_server_for_sap_applications_12_sp3_for_the_aws_cloud_setup_guide.pdf?_ga=2.67759635.36618273.1580225010-1421744106.1546416539

PD: I have added the `-` character to the jinja templates to avoid the new lines creation
https://jinja.palletsprojects.com/en/master/templates/#whitespace-control

### DIFF
--- a/templates/scale_up_resources.j2
+++ b/templates/scale_up_resources.j2
@@ -1,7 +1,7 @@
-{% set data = pillar.cluster.configure.template.parameters %}
-{% set sid = data.sid.upper() %}
-{% set instance = '{:0>2}'.format(data.instance) %}
-{% set cloud_provider = grains['cloud_provider'] %}
+{%- set data = pillar.cluster.configure.template.parameters %}
+{%- set sid = data.sid.upper() %}
+{%- set instance = '{:0>2}'.format(data.instance) %}
+{%- set cloud_provider = grains['cloud_provider'] %}
 
 #
 # defaults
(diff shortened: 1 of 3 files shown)


templates/scale_up_resources.j2
```diff
@@ -14,17 +14,57 @@ rsc_defaults \
...
+
+{%- elif cloud_provider == "microsoft-azure" %}
+
+primitive rsc_socat_{{ sid }}_HDB{{ instance }} anything \
+    params binfile="/usr/bin/socat" cmdline_options="-U TCP-LISTEN:625{{ instance }},backlog=10,fork,reuseaddr /dev/null" \
+    op monitor timeout="20" interval="10" depth="0"
+
+group g_ip_{{ sid }}_HDB{{ instance }} rsc_ip_{{ sid }}_HDB{{ instance }} rsc_socat_{{ sid }}_HDB{{ instance }}
+
+colocation col_saphana_ip_{{ sid }}_HDB{{ instance }} 2000: g_ip_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Master
```
Simranpal: Do we need to group if there is only 1 resource, here rsc_ip

arbulu89: There are 2 resources: `rsc_ip` and `rsc_socat_`

PR #68, opened 2020-02-11
Add prefix prometheus-exporter

MalloZup:


saphanabootstrap-formula.spec
```diff
@@ -19,7 +19,7 @@
 # See also http://en.opensuse.org/openSUSE:Specfile_guidelines
 
 Name:           saphanabootstrap-formula
-Version:        0.4.4
+Version:        0.4.5
```
arbulu89: I think this change deserves a bigger version upgrade. 
I would put `0.5.0` as it looks like it's not backward compatible with hanadb_exporter anymore

saphanabootstrap-formula.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Feb 11 10:51:56 UTC 2020 - Dario Maiocchi <dmaiocchi@suse.com>
+
+-  Version 0.4.5
+  * Change name of pkg to prometheus-hanadb_exporter (jsc#SLE-10545)
```
arbulu89: I would rewrite the message to

`Change the package name to prometheus-hanadb_exporter (jsc#SLE-10545)`
I would prefer to not use SMS kind of texts if we at least find them during the review.

saphanabootstrap-formula.spec
```diff
@@ -31,7 +31,7 @@ BuildArch:      noarch
 Requires:       habootstrap-formula
 Requires:       salt-shaptools
 Requires:       salt-formulas-configuration
-Suggests:       hanadb_exporter >= 0.5.0
+Suggests:       prometheus-hanadb_exporter >= 0.5.0
```
arbulu89: Align this version requirement with the hanadb_exporter new version please

arbulu89: The new prometheus-hanadb_exporter version is `0.7.0`

arbulu89:
@MalloZup I have merged the hanadb_exporter new PR;
https://github.com/SUSE/hanadb_exporter/pull/59

So, we can merge this when the changes are done.

PR #74, opened 2020-03-10
README: follow branding guidelines for "SUSE"

dmacvicar:
From https://www.suse.com/brandcentral/suse/identity.php

"SUSE Corporate NAME
SUSE should always be spelled with all caps. Never lowercase the U (SuSE) or the entire name (suse)."


MalloZup:
thx@

PR #75, opened 2020-03-20
Feature/install pydbapi

arbulu89:
Right now, the hanadb_exporter is using PyHDB to connect to the HANA database. PyHDB is not an officially supported package.
Python dbapi is officially provided by SAP, so this change installs this package.

The change adds an optional new variable: `hana_client_path`
This parameter is used in the `exporter` field, if the user wants to provide a specific HANA version. Remember that the HANA client is not always available in the `software_path` parameter, as this can only store the database.

PD: In this occasion we don't need to reload the hdb_connector package as in sapnwbootstrap-formula, as the python package is used by the exporter, so not during the installation process

### DIFF
--- a/hana/install_pydbapi.sls
+++ b/hana/install_pydbapi.sls
@@ -19,7 +19,7 @@ hana_install_python_pip:
 hana_extract_pydbapi_client:
   hana.pydbapi_extracted:
     - name: PYDBAPI.TGZ
-    - software_folders: [{{ node.install.software_path }}]
+    - software_folders: [{{ node.exporter.hana_client_path|default(node.install.software_path) }}]
     - output_dir: {{ pydbapi_output_dir }}
     - hana_version: '20'
     - force: true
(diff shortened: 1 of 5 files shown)


hana/install_pydbapi.sls
```diff
@@ -0,0 +1,46 @@
...
+    - name: python-pip
+    {% else %}
+    - name: python3-pip
+    {% endif %}
+    - retry:
+        attempts: 3
+        interval: 15
+    - resolve_capabilities: true
+
+hana_extract_pydbapi_client:
```
diegoakechi: @arbulu89 it has the same code as the NW formula. Would it make more sense to move to a generic state on salt-shaptools and use it on both?

arbulu89: What do you exactly mean? The whole sls file?
I thought about that, but I prefer to have the logic visible. I don't want to install and run things like `pip` hidden in some module.
It doesn't require much more maintenance, and by now, the code duplication is minimal.

PR #76, opened 2020-03-26
Feature/sapcar support 

Simranpal:
Adding the sapcar extraction functionality to extract HANA database sar file.
With this PR, I allow did following changes to how HANA installation works.

- Set a Default extraction directory for hana: `/sapmedia/HANA`

- Added a global `software_path` variable, The precedence for hana install media now looks like this: `local node entry for software_path` > `global software_path` > `hdbserver_extract_dir`
I have added in pillar notes that `hdbserver_extract_dir` will be used if both `software_path` are not provided

- Copy the `SIGNATURE.SMF` to  directory containing `hdblcm` executable (ie `SAP_HANA_DATABASE`)

### DIFF
--- a/hana/extract_hdbserver.sls
+++ b/hana/extract_hdbserver.sls
@@ -1,6 +1,6 @@
 {%- from "hana/map.jinja" import hana with context -%}
 
-{% if hana.hdbserver_sar_file and hana.sapcar_exe_file is defined %}
+{% if hana.hdbserver_sar_file is defined and hana.sapcar_exe_file is defined %}
 
 extract_hdbserver_sar:
     sapcar.extracted:
--- a/hana/extract_hdbserver.sls
+++ b/hana/extract_hdbserver.sls
@@ -1,0 +1,22 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
+
+{% if hana.hdbserver_sar_file and hana.sapcar_exe_file is defined %}
+
+extract_hdbserver_sar_{{ host }}:
+    sapcar.extracted:
+    - name: {{ hana.hdbserver_sar_file }}
+    - sapcar_exe: {{ hana.sapcar_exe_file }}
+    - output_dir: {{ hana.hdbserver_extract_dir }}
+    - options: "-manifest SIGNATURE.SMF"
+
+copy_signature_file_to_installer_dir_{{ host }}:
+    file.copy:
+    - source: {{ hana.hdbserver_extract_dir }}/SIGNATURE.SMF
+    - name: {{ hana.hdbserver_extract_dir }}/SAP_HANA_DATABASE/SIGNATURE.SMF
+    - preserve: True
+    - force: True
+    - require:
+        - extract_hdbserver_sar_{{ host }}
+
+{% endif %}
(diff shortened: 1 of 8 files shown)


Simranpal:
Hi @arbulu89 with this PR, the pydbapi_client `software_folders` need to be changed as well I guess https://github.com/SUSE/saphanabootstrap-formula/blob/e9ec20904ac23acf978a8e53e8854b31f994959c/hana/install_pydbapi.sls#L22

hana/extract_hdbserver.sls
```diff
@@ -0,0 +1,22 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
```
arbulu89: We don't need this `host` variable, right?
If we can, I would remove it to make everything simpler

hana/install.sls
```diff
@@ -6,12 +6,18 @@ include:
 
 {% for node in hana.nodes if node.host == host and node.install is defined %}
 
+{% if hana.software_path is defined or node.install.software_path is defined  %}
+{% set software_path =  node.install.software_path|default(hana.software_path) %}
```
arbulu89: Put only 2 spaces after `=`
(the same in the next line)

pillar.example
```diff
@@ -8,14 +8,26 @@ hana:
...
   # Warning: only a unique solution can exist into a node.
   saptune_solution: 'HANA'
+  
+  # Specify the path to the sapcar executable & HANA database server installation sar archive
+  # The sar archive will be extracted to path specified at hdbserver_extract_dir (optional, by default /sapmedia/HANA)
+  # Make sure to use the latest/compatible version of sapcar executable, otherwise file may be extracted incorrectly
+  sapcar_exe_file: '/sapmedia/sapcar.exe'
+  hdbserver_sar_file: '/sapmedia/IMDB_SERVER_LINUX.SAR'
+  hdbserver_extract_dir: '/sapmedia/HANA'
+  # Or specify the path to already extracted HANA platform installation media
```
arbulu89: I would put the `software_path` at the top of installation variables, and comment that this variable has prevalence over `hdbserver` variables

hana/install.sls
```diff
@@ -6,12 +6,18 @@ include:
 
 {% for node in hana.nodes if node.host == host and node.install is defined %}
 
+{% if hana.software_path is defined or node.install.software_path is defined  %}
```
arbulu89: I'm just wondering if the next works
```
{% set software_path = node.install.software_path|default(hana.software_path)|default(hana.hdbserver_extract_dir) %}
```

hana/install_pydbapi.sls
```diff
@@ -16,10 +16,20 @@ hana_install_python_pip:
         interval: 15
     - resolve_capabilities: true
 
+# The software_folders is retrieved in this order: 
+# 1. node.exporter.hana_client_path
+# 2. node.install.software_path
+# 3. hana.software_path
+{% if node.exporter.hana_client_path is defined or node.install.software_path is defined  %}
```
arbulu89: Don't forget the `hana.software_path`. Maybe this works:

```
{% set software_path = node.exporter.hana_client_path|default(node.install.software_path)|default(hana.software_path)|default(hana.hdbserver_extract_dir) %}
```

hana/install.sls
Simranpal: Works as expected

hana/install_pydbapi.sls
Simranpal: hdbserver_extract_dir will contain the HANA Client package so i will  skip adding that

PR #77, opened 2020-03-27
Feature/gcp instance

arbulu89:
Improve `fence_gce` usage using the `gcp_instance_id` created in: https://github.com/SUSE/habootstrap-formula/pull/46

Right now, we were assuming that the hostname is the same as the instance name. This is true in most of the cases, but the hostname can be customized as well, so this change makes the usage always correct.

Besides that, I have remove useless `action` and `pcmk_host_map`. They are not needed.

### DIFF
--- a/templates/scale_up_resources.j2
+++ b/templates/scale_up_resources.j2
@@ -52,7 +52,7 @@ colocation col_saphana_ip_{{ sid }}_HDB{{ instance }} 2000: rsc_aws_vip_{{ sid }
 
 # This stonith resource will be duplicated for each node in the cluster
 primitive rsc_gcp_stonith_{{ sid }}_HDB{{ instance }}_{{ grains['host'] }} stonith:fence_gce \
-    params plug={{ grains['gcp_instance_id'] }} \
+    params plug={{ grains['gcp_instance_id'] }} pcmk_host_map="{{ grains['host'] }}:{{ grains['gcp_instance_id'] }}" \
     meta target-role=Started
 
 primitive rsc_gcp_vip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:gcp-vpc-move-route \
--- a/templates/scale_up_resources.j2
+++ b/templates/scale_up_resources.j2
@@ -52,7 +52,7 @@ colocation col_saphana_ip_{{ sid }}_HDB{{ instance }} 2000: rsc_aws_vip_{{ sid }
 
 # This stonith resource will be duplicated for each node in the cluster
 primitive rsc_gcp_stonith_{{ sid }}_HDB{{ instance }}_{{ grains['host'] }} stonith:fence_gce \
-    params action=reboot plug={{ grains['host'] }} pcmk_host_map="{{ grains['host'] }}:{{ grains['host'] }}" \
+    params plug={{ grains['gcp_instance_id'] }} \
     meta target-role=Started
 
 primitive rsc_gcp_vip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:gcp-vpc-move-route \
--- a/saphanabootstrap-formula.changes
+++ b/saphanabootstrap-formula.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Mar 27 11:16:16 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
+
+- Version 0.5.3
+  * Update the fence_gce usage to use gcp_instance_id 
+
+-------------------------------------------------------------------
 Fri Mar 20 13:19:03 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
 
 - Version 0.5.2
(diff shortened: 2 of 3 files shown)


templates/scale_up_resources.j2
```diff
@@ -52,7 +52,7 @@ colocation col_saphana_ip_{{ sid }}_HDB{{ instance }} 2000: rsc_aws_vip_{{ sid }
 
 # This stonith resource will be duplicated for each node in the cluster
 primitive rsc_gcp_stonith_{{ sid }}_HDB{{ instance }}_{{ grains['host'] }} stonith:fence_gce \
-    params action=reboot plug={{ grains['host'] }} pcmk_host_map="{{ grains['host'] }}:{{ grains['host'] }}" \
+    params plug={{ grains['gcp_instance_id'] }} \
```
gao-yan: If every node in the cluster has the same permission accessing the cloud APIs, we could use a single stonith resource to target multiple nodes. For example:

```
primitive rsc_gcp_stonith stonith:fence_gce \
    params pcmk_host_map="node_name1:gcp_instance_id_1;node_name2:gcp_instance_id2"   
```

Pacemaker will know this stonith resource handles fencing that targets either the cluster node `node_name1` or `node_name2`. It'll map the node name correspondingly to `gcp_instance_id1` or `gcp_instance_id2`, and pass it to the fencing agent `fence_gce` when executing it.  So we could and should also omit `plug`/`port` parameter in the resource configuration.

We don't need to care about where it's shown to be running. It will be registered on all nodes and each node will be able to handle fencing requests.

arbulu89: Hi @gao-yan ,
As discussed externally, this option doesn't work right now. Here why:
The `fence_gce` agent uses `plug` internally to find the node to fence, but the `pcmk_host_map` doesn't map `port` to `plug`, so this variable is not available.

https://github.com/ClusterLabs/fence-agents/blob/master/agents/gce/fence_gce.py#L244

For this reason, by now, we don't have any other alternative than having 2 different stonith resources.

As the reboot operation is available in both nodes, we don't really need to add any colocation constrain.

So, as it is, it should work.

Do you agree @gao-yan ?

gao-yan: Hrm, according to @arbulu89, `fence_gce` requires `--plug` option being supplied and doesn't accept `--port` as a compatible option.

That means for now the stonith resources need `plug` parameter being explicitly configured, otherwise pacemaker will only pass the gcp_instance_id to `--port` option when executing `fence_gce`, not to `--plug`.

This could be improved/fixed in both pacemaker and probably also `fence_gce` agent:

-  Pacemaker should pass the corresponding gcp_instance_id parsed from `pcmk_host_map`  as well to `--plug` of `fence_gce` when executing it.

- Optionally/preferable, `fence_gce` should probably properly support the traditional `--port` parameter as claimed in its metadata although it's marked as deprecated.

For now we have to use multiple stonith resources with specific `plug` explicitly configured and keep the `pcmk_host_map` parameter so that pacemaker knows which cluster node a stonith resource targets.

gao-yan: > For this reason, by now, we don't have any other alternative than having 2 different stonith resources.

True.

> 
> As the reboot operation is available in both nodes, we don't really need to add any colocation constrain.

We don't actually need any constraints for them.

> 
> So, as it is, it should work.

We should keep `pcmk_host_map` so that pacemaker knows which node it targets. So it should work this way:
```
params plug={{ grains['gcp_instance_id'] }} pcmk_host_map="{{ grains['host'] }}:{{ grains['gcp_instance_id'] }}" \
```

gao-yan: FYI,  improvements for handling of fence-agents parameters in pacemaker has been merged upstream:

https://github.com/ClusterLabs/pacemaker/pull/2036

With that:
- Pacemaker handles and passes both `port` and `plug` parameters with the same value if neither of them is explicitly set in resource configuration. 
- Besides `port`, `plug` and `action` are now also shown as `required="0"` in the metadata output by `stonith_admin --metadata` or `crm_resource --show-metadata` (where `crm ra info` is from), so that users know they don't have to configure them. Pacemaker handles and passes the parameters to the agent when executing it.

> If every node in the cluster has the same permission accessing the cloud APIs, we could use a single stonith resource to target multiple nodes. For example:
> 
> ```
> primitive rsc_gcp_stonith stonith:fence_gce \
>     params pcmk_host_map="node_name1:gcp_instance_id_1;node_name2:gcp_instance_id2"   
> ```

@arbulu89 According to the latest experiment we did, this single stonith resource targeting multiple nodes should work even without the above pacemaker changes, no? Since pacemaker has been always handling and passing `port` parameter.

arbulu89: Hi @gao-yan . No, without the change a single resource scenario doesn't work. The `fence_gce` agent complains that `plug` is missing, as the `pcmk_host_map` is not passing the parameter with this value, apparently it only works when `port` or `plug` are used, that makes the need to have 2 resources.
If you have the `rpm` package of this new version, we could test it easily.
We should create a ticket in our internal backlog to test, and if it works, to adapt the template when the new changes reach official channels.

PR #78, opened 2020-03-31
Feature/sr hook

arbulu89:
New HANA database SAPHanaSR hook implementation.
It's pretty simple.
1. Stop HANA
2. Add srhook in global file
3. Add operation mode in global file (only primary)
4. Update sudoers to allow some basic crm operations to sidadm user
5. Start HANA

Guide in (**Chapter 8**)
https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-12/?_ga=2.59467309.1005826984.1585052809-1690579492.1558969243#_operating_system_setup

### DIFF
--- a/hana/ha_cluster.sls
+++ b/hana/ha_cluster.sls
@@ -1,0 +1,89 @@
+{%- from "hana/map.jinja" import hana with context -%}
+{% set host = grains['host'] %}
+
+{% for node in hana.nodes if node.host == host %}
+
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set sap_instance = '{}_{}'.format(node.sid, instance) %}
+
+# Stop SAP HanaError
+stop_hana_{{ sap_instance }}:
+  module.run:
+    - hana.stop:
+      - sid: {{ node.sid }}
+      - inst: {{ node.instance }}
+      - password: {{ node.password }}
+
+# Add SAPHANASR hook
+# It would be better to get the text from /usr/share/SAPHanaSR/samples/global.ini
+configure_ha_hook_{{ sap_instance }}:
+  file.append:
+    - name:  /hana/shared/{{ node.sid.upper() }}/global/hdb/custom/config/global.ini
+    - text: |
+
+        [ha_dr_provider_SAPHanaSR]
+        provider = SAPHanaSR
+        path = /usr/share/SAPHanaSR
+        execution_order = 1
+
+        [trace]
+        ha_dr_saphanasr = info
+    - require:
+      - stop_hana_{{ sap_instance }}
+
+# Configure system replication operation mode in the primary site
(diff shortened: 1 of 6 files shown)


hana/ha_cluster.sls
```diff
@@ -0,0 +1,89 @@
...
+sudoers_edit_{{ sap_instance }}:
+  file.copy:
+    - name: {{ sudoers }}
+    - source: {{ tmp_sudoers }}
+    - force: true
+    - require:
+      - sudoers_check_{{ sap_instance }}
+      - stop_hana_{{ sap_instance }}
+
+# Stop SAP HanaError
```
diegoakechi: @arbulu89 I think the comment here should be start instead.

hana/ha_cluster.sls
```diff
@@ -0,0 +1,89 @@
...
+        ha_dr_saphanasr = info
+    - require:
+      - stop_hana_{{ sap_instance }}
+
+# Configure system replication operation mode in the primary site
+{% for secondary_node in hana.nodes if node.primary is defined and secondary_node.secondary is defined and secondary_node.secondary.remote_host == host %}
+configure_replication_{{ sap_instance }}:
+  file.line:
+    - name:  /hana/shared/{{ node.sid.upper() }}/global/hdb/custom/config/global.ini
+    - content: operation_mode = {{ secondary_node.secondary.operation_mode }}
```
Simranpal: Hi @arbulu89 Changing the HANA operation mode is not really straightforward. Changing to `logreplay` is fine. But normally when we switch HANA's operating mode to `delta_datashipping` it requires us to do a full data shipping i.e. get a copy of full data from HANA primary node. This is done by running the secondary registration command with `--force_full_replica ` parameter.

https://help.sap.com/viewer/6b94445c94ae495c83a19646e7c3fd56/2.0.02/en-US/627bd11e86c84ec2b9fcdf585d24011c.html

https://launchpad.support.sap.com/#/notes/0002500677

diegoakechi: Actually, we already configure the operation mode when establishing the system replication via the https://github.com/SUSE/saphanabootstrap-formula/blob/master/pillar.example#L82, so we are not changing the operation mode. The point is if it is really necessary to persist it on the global.ini or if HANA already does it during the SR enablement...

arbulu89: @Simranpal @diegoakechi This is what the recently edited guide says hehe
If we have some issue with that, we should contact with them too.

diegoakechi: I am fine with it since it will be executed in order here, so it will never change the operation mode inside the same execution.

Simranpal: Yes @diegoakechi is correct. We will not encounter the issue I had in mind (changing from `logreplay` to `delta_datashipping`). So this works fine.
The mode is set during SR setup and updating the global.ini file with same operation mode should not cause any effect.

PR #81, opened 2020-04-21
Feature/adjust cost optimized

Simranpal:
Adjusting the `py_packages_folder` used for cost optimized required packages `pydbapi.tgz` and `hdbcli` archive.
Previously grains entry `hana_inst_folder` was used, which may not be ideal to use in formula. I have updated it to fetch the client packages from `software_path` variable.

Another solution is to update our salt function `pydbapi_extracted` & consider passing the extra options there for extracting pydbapi archive. However, as the structure for pydbapi varies for HANA versions, as well the hdbcli package versions inside have different structure across HANA revisions, we may still need to use the current salt `archive.extracted`

Tested these changes with cost optimized deployment and works as expected. I also tested the existing formula and it was working fine as well

### DIFF
--- a/hana/enable_cost_optimized.sls
+++ b/hana/enable_cost_optimized.sls
@@ -50,7 +50,8 @@ failure:
     - failhard: True
 {% endif %}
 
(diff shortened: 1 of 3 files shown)


Simranpal:
Hi @arbulu89 this is more of temporary fix. I will work in changing the autodetection methods next to support the case, when hdbserver sar file is provided. We can merge this for meantime

PR #82, opened 2020-05-05
Exporter Pacemaker resource

stefanotorresi:
This PR implements the Pacemaker resource to migrate `hanadb_exporter` from one node to another when a node failure happens.

It refactors the salt state slightly to make it more consistent with all the other formulas, e.g. by renaming it to `monitoring.sls` and using the same flag as in the other formulas.

Integrated in SUSE/ha-sap-terraform-deployments#465.


pillar.example
```diff
@@ -23,14 +23,19 @@ hana:
   # To use this option the primary and secondary nodes must be defined in the pillar file
   ha_enabled: true
 
+  # optional: enables monitoring via ha_cluster_exporter (disabled by default)
```
arbulu89, line 26: Change `ha_cluster_exporter` by `hanadb_exporter`

hana/monitoring.sls
```diff
@@ -0,0 +1,54 @@
...
+      - prometheus-hanadb_exporter
+
+hanadb_exporter_metrics_configuration:
+  file.managed:
+    - name: /etc/hanadb_exporter/metrics.json
+    - source: /usr/etc/hanadb_exporter/metrics.json
+    - require:
+      - prometheus-hanadb_exporter
+
+{% for node in hana.nodes if node.host == grains['host'] %}
```
arbulu89, line 26: I think we need to change the strategy here. We cannot just copy the configuration file in all the nodes and start the exporter, as some of them don't have the `exporter` data in the pillar file, and the template cannot be rendered. In fact, some of the entries cannot have the exporter running (the secondary machine for example).

This is what I would do:
- Create the config file only the nodes where the `exporter` entry exist
- Start the exporter only for the previous condition too
- We need to replicate this config file in any other node that shares the `sid` and `instance number`, but without starting the exporter. We might only want this if `ha_enabled` param is true.

Have in mind that when we export metrics from a HANA env with system replication, we only need metrics from the primary, and not the secondary. We need to play with this.

stefanotorresi: The problem I have is that `hana.nodes` doesn't actually model the nodes, but it models HANA instances.
So in the example deployment, for instance, we have 3 elements in the pillar, but only 2 actual nodes.
If I try to replicate the primary config in all the nodes, I would get a third attempt at creating the same configuration in node 2, because of the third element in the array for the QA instance.
Should we rethink this completely or find a way around it?

arbulu89: That's why I suggested to copy the configuration files only if the `sid` and `instance number` match. It's not possible to have 2 sap instances in the same node with that data, so we shouldn't never have the scenario you comment.

For example:
- I have 3 sap instances defined as in our example.
- I only define the exporter in the first primary node (and optionally in the QAS, this depends on the user desires)
- I don't need to define the exporter data in the secondary, as we cannot start the exporter there
- In this case only first node sid/instance number and secondary sid/instance number match, so we would create a copy of the exporter config file there. The state would be executed only twice, it's not possible to have duplicated things here

hana/monitoring.sls
```diff
@@ -0,0 +1,104 @@
...
+      - prometheus-hanadb_exporter
+      - hanadb_exporter_metrics_configuration
+      - hanadb_exporter_logging_configuration
+    - context:
+        host: {{ node.host }}
+        exporter: {{ exporter|yaml }}
+        sap_instance_nr: "{{ sap_instance_nr }}"
+        exporter_instance: {{ exporter_instance }}
+
+{% if node.primary is defined or node.secondary is defined %}
```
arbulu89, line 88: This condition doesn't make sense to me.
What if I want to monitor a system replication that is not using HA?
This is conditioned by the `ha_enabled` flag, no if it's primary or secondary.
In fact, I would enable always for non secondary nodes, and disable only for secondary nodes

templates/hanadb_exporter.j2
```diff
@@ -1,30 +1,20 @@
...
   "hana": {
-    "host": "{{ node.host }}",
-    {% if node.exporter.multi_tenant|default(true) is sameas true and node.exporter.port is not defined %}
-    "port": 3{{ dbinst }}13,
-    {% else %}
-    "port": {{ node.exporter.port }},
-    {% endif %}
-    "user": "{{ node.exporter.user }}",
-    "password": "{{ node.exporter.password }}"
+    "host": "{{ grains['host'] }}",
```
arbulu89, line 7: This should be just `{{ host }}`, right?
This is why we are setting the value in the context `hanadb_exporter_configuration_{{ exporter_instance }}`

hana/monitoring.sls
```diff
@@ -0,0 +1,104 @@
+{%- from "hana/map.jinja" import hana with context -%}
+
+{% set pydbapi_output_dir = '/tmp/pydbapi' %}
+
+{% for node in hana.nodes if node.host == grains['host'] %}
+
+{% if node.secondary is not defined %}
+{% set exporter = node.exporter %}
```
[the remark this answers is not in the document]
stefanotorresi: Hmm, this is a problem that is present everywhere in these salt formulas: we cannot validate group of fields so that they make logical sense together.

In the other formulas we don't have all this granularity: if we enable monitoring it just works, we don't need to specify anything else to configure the exporters correctly, it's done by the state. That said, the HANA exporter is a bit different and more complicated, as we discussed already, so I guess that approach doesn't apply here.

That said, we could mitigate this by putting all of the rest of the salt state under an `if` condition on the `exporter` variable not being `null`, similarly to how it was being done before.

templates/hanadb_exporter.j2
stefanotorresi: actually `host` will always be equal to `grains['host']`; I forgot to remove the additional variable from the context.

arbulu89: Ok, remove this or to other, but both shouldn't be there hehe

PR #83, opened 2020-05-09
Update cost optimized state to use pydbapi_extracted functionality

Simranpal:
With this PR the cost optimized state will use our custom `pydbapi_extracted` function instead of salt's `archive.extracted` to utilize the automatic detection / finding hana software folders within sap media

Also removed the hardcoded client package path and now it uses the software path for pyadbapi location

To-Do:
-Add support for HANA 1.0 Client which has different folder structure. It has the client files inside `hdcli` folder inside the `pydbapi` tar vs `hdcli-version-tar`
-Test with deployment scenarios
-Update spec, change log

Related to https://github.com/SUSE/salt-shaptools/pull/61


hana/enable_cost_optimized.sls
```diff
@@ -50,15 +50,14 @@ failure:
...
-
 extract_hana_pydbapi_archive:
-    archive.extracted:
-      - name: /hana/shared/srHook
-      - enforce_toplevel: False
-      - source: {{ py_packages_folder }}
-      - options: --transform 's|-[0-9]*\.[0-9]*\.[0-9]*|-package|' --wildcards 'hdbcli*'
+    hana.pydbapi_extracted:
+      - name: PYDBAPI.TGZ
+      - software_folders: [{{ hana.software_path }}]
```
arbulu89, line 56: I think we need to put `node.install.software_path|default(hana.software_path)` here

Simranpal: Thanks, I updated it

arbulu89:
@Simranpal I think we can postpone the `HANA 1.0` implementation by now. What do you think?
Eventually, if `HANA 1.0` is required (we need to discuss this), we will do all the needed changes. I think there are more than only this.

Besides that, we only need to do the other things you wrote in the main description.

Simranpal:
Thanks @arbulu89 as discussed this works for `HANA 2.0` I did the testing with cost opt deployment and works as expected

arbulu89:
> Thanks @arbulu89 as discussed this works for `HANA 2.0` I did the testing with cost opt deployment and works as expected

Thanks @Simranpal , so update the spec and changes file and we can merge

PR #84, opened 2020-05-14
Add support to extract zip,rar,exe,sar hana media

Simranpal:
SAP HANA installation media is available in these formats: `ZIP, RAR, SAR and Multipart exe RAR`.

This PR brings the feature to automatically detect the HANA archive type and extract the archive accordingly.
Now, there will be one variable `hana_archive_file` which will take the media archive package as input.
Please suggest if better names for variables.

For detecting the file extension type salt does not have such feature, so I have added the check in jinja conditions. I tested the extractions for all archive types and works as expected.

To-Do:
-Test with deployment scenarios
-Update the spec, change log


hana/extract_hana_package.sls
```diff
@@ -0,0 +1,47 @@
...
+    archive.extracted:
+      - name: {{ hana_extract_dest }}
+      - enforce_toplevel: False
+      - source: {{ hana_package }}
+
+{% elif (".exe" in hana_package) or (".EXE" in hana_package) %}
+
+extract_hana_multipart_archive:
+  cmd.run:
+    - name: unrar x {{ hana_package }}
```
arbulu89, line 25: We should install `unrar` before running this method. It might always come, but maybe it's not installed by default in all images. It looks a good check

hana/extract_hana_package.sls
```diff
@@ -0,0 +1,47 @@
+{%- from "hana/map.jinja" import hana with context -%}
+
+{% if hana.hana_archive_file is defined %}
+{% set hana_package = hana.hana_archive_file %}
+{% set hana_extract_dest = hana.hana_extract_dir %}
```
arbulu89, line 5: Why have you changed from `hana_extract_dir` to `hana_extract_dest`? If you prefer, change the default name in `defaults.yml` and pillar file, now that you have changed. Put the best one from the beginning.
I would just keep the same name to make the code easier to read.
The same for `hana_packages`.
It's only cosmetics, and It's not a big deal, just curiosity.

hana/extract_hana_package.sls
```diff
@@ -0,0 +1,47 @@
...
+
+setup_hana_extract_directory:
+    file.directory:
+      - name: {{ hana_extract_dest }}
+      - mode: 755
+      - makedirs: True
+
+{% if (".ZIP" in hana_package) or (".zip" in hana_package) or (".RAR" in hana_package) or (".rar" in hana_package) %}
```
arbulu89, line 13: Maybe better:
```{%- if hana_package.endswith((".ZIP", ".zip", ".RAR", ".rar")) %}```
What do you think?

File: hana/extract_hana_package.sls [hunk 4]
```diff
@@ -0,0 +1,47 @@
...
+
+extract_hana_archive:
+    archive.extracted:
+      - name: {{ hana_extract_dest }}
+      - enforce_toplevel: False
+      - source: {{ hana_package }}
+
+{% elif (".exe" in hana_package) or (".EXE" in hana_package) %}
```
arbulu89, line 21: {%- if hana_package.endswith((".exe", ".EXE")) %}

hana/extract_hana_package.sls
```diff
@@ -0,0 +1,47 @@
...
+
+extract_hana_multipart_archive:
+  cmd.run:
+    - name: unrar x {{ hana_package }}
+    - cwd: {{ hana_extract_dest }}
+
+{% elif ((".sar" in hana_package) or (".SAR" in hana_package)) and hana.sapcar_exe_file is defined %}
```
arbulu89, line 28: {%- if hana_package.endswith((".sar", ".SAR")) and hana.sapcar_exe_file is defined %}

hana/extract_hana_package.sls
Simranpal: I think `hana_extract_dest` is better and aligns with the `dest` parameter for extraction function s in salt, unix. I will use it for all locations.

hana/extract_hana_package.sls
Simranpal: Yes thats better, I tested it and works fine

hana/extract_hana_package.sls
Simranpal: any reason you put `if` instead of `elif`? With if I would have to also add `endif` after each condition. With elif looks cleaner I think

Simranpal:
I am testing with the deployments and will update once its successful

hana/extract_hana_package.sls
arbulu89: Yes, the `elif` is fine. I just copy/pasted the previous command to add the `endswith` option

Simranpal:
> @Simranpal I noticed that in `sapnwbootstrap-formula` we use `swpm_extract_dir`.
> In order to be aligned, I would let this to `hana_extract_dir`, or change the other.
> I think by now, we can let `hana_extract_dir` to avoid creating other change in the other project.
> What do you think?

@arbulu89 it was named as `hdbserver_extract_dir` for sar extract earlier, and since now we are extracting all hana packages, we need to rename it better I think, `hana_extract_dest` or `hana_extract_dir` any is fine.

For the NW part now there are other nw packages extractions, the `swpm_extract_dir` may need to be removed or changed. I am thinking there to have one `netweaver_extract_dest` and put the swpm, kernel, export, other packages extracts there if archives provided.

So I will set the variable name `hana_extract_dir` here and updated the NW formula variables accordingly.

hana/extract_hana_package.sls
Simranpal: Thinking again, I prefer to use `hana_extract_dir` because this variable is used for extraction and later for installation, and for installation having `dest` word as input variable is confusing

diegoakechi:
@arbulu89 I think it is safe to use the wrapper to avoid `ifs` giving the different syntax. This wrapper is officially documented on the SLE-15 release notes:

https://www.suse.com/releasenotes/x86_64/SUSE-SLES/15/#fate-323896

Important to notice that AFAIK the package unrar was available on SLE-12, and the wrapper doesn't exist there, so better test and adjust the code for this case, as it will probably not work on SLE-12 versions.

arbulu89:
> @arbulu89 I think it is safe to use the wrapper to avoid `ifs` giving the different syntax. This wrapper is officially documented on the SLE-15 release notes:
> 
> https://www.suse.com/releasenotes/x86_64/SUSE-SLES/15/#fate-323896
> 
> Important to notice that AFAIK the package unrar was available on SLE-12, and the wrapper doesn't exist there, so better test and adjust the code for this case, as it will probably not work on SLE-12 versions.

Thanks @diegoakechi 
So this is important, we want to make the code available for SLE12 codestream too, @Simranpal could you test this please?

Simranpal:
> Thanks @diegoakechi
> So this is important, we want to make the code available for SLE12 codestream too, @Simranpal could you test this please?

Thanks @diegoakechi @arbulu89 testing with SLE12 did not work as expected. So I have updated the salt code to check SLE version and install the necessary unrar package

saphanabootstrap-formula.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Wed Jun 10 01:49:02 UTC 2020 - Simranpal Singh <simranpal.singh@suse.com>
+
+- Add support to extract zip,rar,exe,sar hana media
```
arbulu89, line 4: Please, comment that this change has non backward compatible changes (`hdbserver_extract_dir` changed by `hana_extract_dir`)

Simranpal:
Hi @arbulu89 I have some further changes, to fix the multipart exe archive extraction and usage in other places like monitoring, cost-optimized. If this looks good now I can merge

PR #91, opened 2020-06-25
Update the saltstack repo url for build validation

Simranpal:
Updating the source configuration for salt gpg key in travis build configuration.
It seems that the saltstack repo removed the Ubuntu `14.04` distrution, and we are getting build error in our formulas due to using the old url.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ services:
 addons:
   apt:
     sources:
-      - sourceline: 'deb http://repo.saltstack.com/apt/ubuntu/14.04/amd64/latest trusty main'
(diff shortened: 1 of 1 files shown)


diegoakechi:
@arbulu89 AFAIK end of July.  @paususe can you confirm here?

paususe:
> @arbulu89 AFAIK end of July. @paususe can you confirm here?

@diegoakechi that's right, we will release Salt 3000 (+ patches) on 21.07.2020 for SLES 12 (all SPs) and SLES 15 (all SPs)

BTW, Ubuntu 20.04 removed Salt from their repositories, so now the only way to get Salt on Ubuntu is either SaltStack packages, or SUSE Manager / Uyuni client tools packages.

PR #97, opened 2020-09-03
Fix pillar.example file location in package

Simranpal:
Changing the location of `pillar.example` file in package,
and fix typo in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Find the latest development repositories at  SUSE's Open Build Service[network:h
 
 ```
 git clone https://github.com/SUSE/saphanabootstrap-formula
-cp -R cluster /srv/salt
+cp -R hana /srv/salt
 ```
 
 **Important!** The formulas depends on [salt-shaptools](https://github.com/SUSE/salt-shaptools) package. Make sure it is installed properly if you follow the manual installation (the package can be installed as a RPM package too).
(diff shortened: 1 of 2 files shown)


arbulu89:
@Simranpal How is this PR? I thought we were going to add the `pillar.example` as documentation. So it should go in the `doc` entry, and not in metadata folder.

Simranpal:
> @Simranpal How is this PR? I thought we were going to add the `pillar.example` as documentation. So it should go in the `doc` entry, and not in metadata folder.

@arbulu89 following the other salt formulas , I placed the `pillar.example` in metadata folder. There is no `doc` folder created in other formulas I see.
https://github.com/SUSE/salt-formulas

arbulu89:
> > @Simranpal How is this PR? I thought we were going to add the `pillar.example` as documentation. So it should go in the `doc` entry, and not in metadata folder.
> 
> @arbulu89 following the other salt formulas , I placed the `pillar.example` in metadata folder. There is no `doc` folder created in other formulas I see.
> https://github.com/SUSE/salt-formulas

I don't see any documentation about having the `pillar.example` in the metadata folder, but if everyone else do so...
The `doc` folder is created for all the installed packages (check the `doc` keyword in the spec file).
In SUSE, this folder links to `/usr/share/doc/packages`

arbulu89 (approved):

MalloZup (changes requested):
I'm ok with this. Only can we document this somewhere? 

Can you create maybe some entry on confluence?

MalloZup (approved):

PR #99, opened 2020-09-10
WIP: Make hana installation working without posix attributes

MalloZup:
So after an extensive research and readings on sap documentations etc (https://help.sap.com/doc/e9702d76c3284623b02de196c0e79e49/2.0.05/en-US/SAP_HANA_Server_Installation_Guide_en.pdf) 
I found out that even if the sapys is by default 79, this **doesn't** mean that users/auto-install cannot override it. 
Even if others upstream doc are stating this somehow not clearly.

anyways for us mean we **don't need unix/posix attributes on AD set**
 which is quite good and spare us lot of work and enable cloud integrations.

the pseudocode works ( I hardcoded the uid/id on a real scenario, after stopping states etc)
I might just need to adapt the salt code to get the `id prdadm` getting the ID dynamically and then pass them to the formula, and checking if the node is on ad.

I think for netweaver the same mechanism should exist for overriding this default parameters (didn;'t checked)

### DIFF
--- a/hana/install.sls
+++ b/hana/install.sls
@@ -48,4 +48,8 @@ hana_install_{{ node.host+node.sid }}:
       - {{ key }}: {{ value }}
       {% endfor %}
     {% endif %}
+   {% if grains.get('ad_server', True) %}
+   - require:
+      - cmd: add_sidadm_grains
+   {% endif %}
 {% endfor %}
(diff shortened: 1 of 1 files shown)


MalloZup:
closing since removed

PR #103, opened 2020-09-22
Use /usr/etc by default

MalloZup:
# DESC:

this pr change to use `/usr/etc` by default for configuration files.

I do think this is safe enough for all systems. All linux systems have that folder allthough isn't used.

https://github.com/SUSE/hanadb_exporter/pull/78

works

### DIFF
--- a/hana/monitoring.sls
+++ b/hana/monitoring.sls
@@ -52,25 +52,12 @@ prometheus-hanadb_exporter:
   - require:
       - install_pydbapi_client
 
-hanadb_exporter_logging_configuration:
-  file.managed:
-    - name: /etc/hanadb_exporter/logging_config.ini
-    - source: /usr/etc/hanadb_exporter/logging_config.ini
-    - require:
-      - prometheus-hanadb_exporter
-
-hanadb_exporter_metrics_configuration:
-  file.managed:
-    - name: /etc/hanadb_exporter/metrics.json
-    - source: /usr/etc/hanadb_exporter/metrics.json
-    - require:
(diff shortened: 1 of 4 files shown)


hana/monitoring.sls
```diff
@@ -52,25 +52,12 @@ prometheus-hanadb_exporter:
...
       - install_pydbapi_client
 
-hanadb_exporter_logging_configuration:
-  file.managed:
-    - name: /etc/hanadb_exporter/logging_config.ini
-    - source: /usr/etc/hanadb_exporter/logging_config.ini
-    - require:
-      - prometheus-hanadb_exporter
-
-hanadb_exporter_metrics_configuration:
```
MalloZup, line 62: we don't need this since `hanadb` exporter can lookup in both folders

arbulu89 (commented):
@MalloZup It looks fine. Anyway, add the new changes in the `changes` file when you have finished.
After testing, I will review back again

MalloZup:
@arbulu89  yes inddeed I forgot the `changes` file :rabbit:

Simranpal (commented):
@MalloZup just question about this location` /usr/etc,` is there any convention or its default we set., when we use that location for storing the exporter configurations?

arbulu89 (approved):
@MalloZup The changes look good! Please, test before merging.

MalloZup:
@Simranpal  yes is a change made by os people https://kubic.opensuse.org/blog/2019-12-05-usr-etc/ it will come on next os

PR #104, opened 2020-09-24
Change sap hana media extraction location 

Simranpal:
To avoid conflict with ready-only sap media mount points, we need to update the location where SAP HANA archives are extracted. Currently, both the default mounting point for sap media, and extraction path for sap media are same: `/sapmedia/HANA`
This can cause issues as mentioned by Peter Schinagl in
https://github.com/SUSE/ha-sap-terraform-deployments/pull/546

1) Could you suggest if there is a better naming for this extraction directory?
2) Based on my research, there is no restriction on where this location can be, except it has to have right permission. 

However, I think placing it under the `/hana` sub directories may not be a good idea. Considering that `/hana/data` `/hana/log`
may run out of disk space, and `/hana/shared/` being used as shared mount directory amongst the hana hosts, I am using a separate directory for hana extraction media.

To-do:
- Test with the installations
- Update the terraform project with this changes
- Follow up with same changes on Netweaver

### DIFF
--- a/hana/defaults.yaml
+++ b/hana/defaults.yaml
@@ -1,6 +1,6 @@
 hana:
   install_packages: true
   ha_enabled: true
-  hana_extract_dir: /sapmedia/HANA
+  hana_extract_dir: /sapmedia_extract/hana
   nodes: []
   monitoring_enabled: false
--- a/hana/defaults.yaml
+++ b/hana/defaults.yaml
@@ -1,6 +1,6 @@
 hana:
   install_packages: true
   ha_enabled: true
-  hana_extract_dir: /sapmedia_extract/hana
+  hana_extract_dir: /sapmedia_extract/HANA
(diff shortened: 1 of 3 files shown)


Simranpal:
cc @petersatsuse

arbulu89 (approved):
@Simranpal It looks fine.
I would put `hana` in capital letters (this is how we have it in the terraform project and NW project).

Besides that, add a comment in the `pillar.example` saying that this should be a new folder that is not the folder where the compressed files are present, to avoid conflicts in file permissions.

Finally, add this change in the changes file. **Please, don't upgrade the version. We will do it when all of the fixes are done**

Simranpal:
I have tested with the deployment and related changes to terraform project done in https://github.com/SUSE/ha-sap-terraform-deployments/pull/538

PR #105, opened 2020-09-28
[Rfc] standalone formula with provided script

MalloZup:
# Workflow from user:

1) User install rpm of formula `zypper in formula-name`

2) User edit `unique` pillar file in `/usr/share/salt-formulas/config/hana/pillar/hana/hana.sls`

This file need to be edited with HELP of suse.documentation which explains formula API etc.
--> having a 1 pillar.example file is normal in saltstack formulas.. see upstream : https://github.com/saltstack-formulas/openldap-formula/blob/master/pillar.example

3)  User use `run-saphanaboostrap-formula` (`run-$FORMULA_NAME`)

# Features of formula CLI:

- colors (red error, green INFO)
- logger and execution phases: 
     * `[PREFLIGHTS]` checks,  ( this can be useful in future , to check if the formula could run in the given host). It should  however stay minimalist because doing to much "defensive" programming is a bad desing (anyways is out of scope)
     * `[SETUP]` copy grains etc ,
     * `FORMULA` run formula..
- set log level `-l` for salt. Default `error`
- `concurrent` and conflict safe. An user can install X formulas, they don't interfer each other. This is because 
`/usr/share/salt-formulas/config/FORMULA_NAME` will contain all self-isolated salt pillars/grains and other config.
There is no shared conf

- minimalism (auto-detection, no input allowed) -> this allow scriptable later on.
    * an admin, could setup 2 pillar files `hana.sls` copy them to host1 and host2
       then run `ssh root@hana1 run-sapformula ; ssh root@hana2 run-sapformula`

# Technical details

For technical details check the RFC: https://confluence.suse.com/pages/viewpage.action?pageId=581566497

The 2 main mechanism are:

* 1) RPM take care to copy the  needed **static** config files`minion` and dir necessary for grant to have 0 conflicts
* 2) After pkg install, the `cli` check we have the folder we needs and execute salt-call with all needed parameters etc

-This is what the CLI for each formula support:

```bash
run-saphanabootstrap-formula -h
Install hana-formula, executing salt-call.

Supported Options:
  -l [LOG_LEVEL]   Set log level of saltstack. Default: error. 
  -h               Show this help.
```

# Possible improvements:

- The description of the CLI might need to be improved  on the help message.
- The RPM packaging could maybe improved, I didn't check the detail there.


config/minion
```diff
@@ -0,0 +1,9 @@
+# we need to have different top file per formulas so call it with postfix formula
+# top-hana.sls , top-netweaver.sls etc etc
+
+state_top: top-hana.sls
```
arbulu89, line 4: I think we don't need the `state_top` or `file_roots` entries here. They can be passed in the `salt-call` execution command

config/pillar/top.sls
```diff
@@ -0,0 +1,3 @@
+base:
+  '*'
+    - hana.hana
```
arbulu89, line 3: Is this correct? This forces the user to have a fixed `hana.sls` pillar file to the user.

arbulu89 (commented):
@MalloZup The idea of having an easy deployment script is good. 
In any case, I think that this is a sensitive topic. SUSE has multiple salt formulas (most of them managed by the SUMA team). The first discussion should be with them. To see if they have a similar need and in that case how they are solving/how the would do it.

About the current implementation.
- I would prefer to have the script working without fixed files. Create all the files during the runtime, and destroy afterwards. This should be easy to do with `bash`
- I would use other approach to get the pillar file. Some flag, so you don't need to have the pillar file in an specific location. `-p my_pillar` or similar.
- The same for the output log file.
- What is the `hana-top.sls` used for? Your comments in the PR description are not clear for me. salt doesn't need 3 `top.sls` files. It needs 2, one for the pillar and the other for the salt execution

config/minion
MalloZup: state_top can't be passed

arbulu89:
@MalloZup By the way, just running `salt-call --local -l info state.sls hana` or `salt-call --local -l info state.apply hana`, we can ignore all the `top.sls` files for the states part. Just needing the pillar files structure

MalloZup:
ok great thx @arbulu89  for feedback

config/minion
MalloZup: resolved as comment

MalloZup:
```NOTE```: I will present this on next sprint-review meeting so no hurry to merge.

Any feedback is welcome. I already tried it out the PKG and worked ok.

**Comment by MalloZup** [comment]
image



this is how it looks like. Note I broke intentionally the install

File: config/pillar/top.sls (hunk 2 above)
Comment (MalloZup) [reply config/pillar/top.sls:3]: Yes I think is good to force to have 1 file as global conf imho. Otherwise I could not spot any valid alternatives

File: run-saphanabootstrap-formula [hunk 3]
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash -e
+
+# Script to install salt formulas 
+
+# these variables are formula specific:
```
MalloZup, line 5: this var change by formula..

stefanotorresi:
Closing due to inactivity.

PR #106, opened 2020-09-30
Use native fencing optionally

arbulu89:
Enable native fencing optionally.
At some cases, `sbd` is used as fencing mechanism, and for those cases, we need to disable the native fencing mechanism.
By now, we only have it for `gcp` and `aws` (for `azure` will be the same).

FYI: @peteratsuse


Simranpal (approved):
Looksgood!

MalloZup (approved):

PR #90, opened 2020-06-19
Update suma form with new entries

Simranpal:
Adding some missing entries for the suma form for HANA. 
I need to validate with Suse Manager yet.

PR #107, opened 2020-10-03
update Suse manager form and fix hana_extract_dir variable usage

Simranpal:
1) Update suma form with new entries mainly related to:
- saptune
- ha cluster states
- hana media extraction logic
- hdb password xml file
- monitoring
- fix for hana installation media preference in the form (this matches the actual usage in pillar )

2) Update how the `hana_extract_dir`  is  imported/used in hana media extraction states
3) Update pre validation logic

This PR replaces https://github.com/SUSE/saphanabootstrap-formula/pull/90

Tested with Suse manager `4.1`  web interface and highstates

### DIFF
--- a/form.yml
+++ b/form.yml
@@ -104,8 +104,14 @@ hana:
         $visibleIf: .install_checkbox == true
         $optional: true
         $type: group
+        local_software_path_checkbox:
+          $name: Use local HANA installation media
+          $type: boolean
+          $default: false
+          $help: Specify the installation media on this node, otherwise global software path will be used.
         software_path:
           $name: Path to local HANA installation media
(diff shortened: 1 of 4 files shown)


arbulu89 (approved):
@Simranpal Please change the changelog with the correct names.
Besides that, the changes look good.
Shouldn't we fix here the fact the instance number is not being rendered correctly and use quotes in the instance number usage?

Simranpal:
> @Simranpal Please change the changelog with the correct names.
> Besides that, the changes look good.
> Shouldn't we fix here the fact the instance number is not being rendered correctly and use quotes in the instance number usage?

Thanks @arbulu89 as discussed, the issue with quotes in the isntance number is not from SUMA but from salt state.

PR #108, opened 2020-10-07
Workaround for installation when using exe archive

Simranpal:
The underlying problem:

Using the `unrar`/ `unar` packages to extract multipart rar archives... did not support/work for extraction with top-level directory skipping.
Saltstack rar extraction functions did not work as well.
For example: `51053381_part1.exe`  `51053381_part2.rar` .. multipart archives will extract to a directory named `51053381` under location we specify, and not directly to specified output directory.
This becomes problem when we want to tell the installer where hana media is located.
The hana installer needs to know this folder name (as part of complete software path) which contains the `LABEL.asc` file.

Workaround:

I did the jinja workaround logic to find /calculate the folder name, from the name of archive itself, and append it to path, which was working fine, until recent change https://github.com/SUSE/saphanabootstrap-formula/commit/5f5b87fcaa2033cf0da7a513d7289b327578f1d4

Current workaround:

I updated the logic in install, monitoring, cost-opt states, to avoid importing variables from extraction sls file(which is problem when using with SUMA highstate)  and moved the  folder name workaround logic to these states.

Tested with Azure deployment using exe archive media.

### DIFF
--- a/saphanabootstrap-formula.spec
+++ b/saphanabootstrap-formula.spec
@@ -36,7 +36,6 @@ Suggests:       prometheus-hanadb_exporter >= 0.7.0
 %define fname hana
 %define fdir  %{_datadir}/salt-formulas
 %define ftemplates templates
-%define fmacros macros
 
 %description
 SAP HANA deployment salt formula. This formula is capable to install
@@ -77,7 +76,6 @@ fi
 
 %attr(0755, root, salt) %{fdir}/states/%{fname}
 %attr(0755, root, salt) %{fdir}/states/%{fname}/%{ftemplates}
-%attr(0755, root, salt) %{fdir}/states/%{fname}/%{fmacros}
 %attr(0755, root, salt) %{fdir}/metadata/%{fname}
 
 %changelog
--- a/saphanabootstrap-formula.spec
+++ b/saphanabootstrap-formula.spec
@@ -36,6 +36,7 @@ Suggests:       prometheus-hanadb_exporter >= 0.7.0
 %define fname hana
 %define fdir  %{_datadir}/salt-formulas
 %define ftemplates templates
+%define fmacros macros
 
 %description
 SAP HANA deployment salt formula. This formula is capable to install
@@ -76,6 +77,7 @@ fi
 
 %attr(0755, root, salt) %{fdir}/states/%{fname}
 %attr(0755, root, salt) %{fdir}/states/%{fname}/%{ftemplates}
+%attr(0755, root, salt) %{fdir}/states/%{fname}/%{fmacros}
 %attr(0755, root, salt) %{fdir}/metadata/%{fname}
 
 %changelog
--- a/hana/enable_cost_optimized.sls
+++ b/hana/enable_cost_optimized.sls
@@ -1,16 +1,8 @@
 {%- from "hana/map.jinja" import hana with context -%}
+{%- from 'hana/macros/get_hana_exe_extract_dir.sls' import get_hana_exe_extract_dir with context %}
+{% set hana_extract_dir = get_hana_exe_extract_dir(hana) %}
 
 {% set host = grains['host'] %}
-{% set hana_extract_dir = hana.hana_extract_dir %}
-
-# Below is temporary workaround to update the software installation path when using unrar for HANA multipart rar archive
-# TODO: Find better solution to set or detect the correct extraction path when extracting multipart rar archive
-# Below logic finds the extraction location based on name of multipart exe archive filename
-{%- if hana.hana_archive_file is defined and hana.hana_archive_file.endswith((".exe", ".EXE")) %}
-{% set archive_base_name = salt['file.basename']( hana.hana_archive_file.split('.')[0]) %}
-{% set archive_name = archive_base_name.split('_')[0] %}
-{% set hana_extract_dir = hana_extract_dir| path_join(archive_name) %}
-{% endif %}
 
 {% for node in hana.nodes %}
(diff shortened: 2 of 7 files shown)


Simranpal:
@arbulu89 @MalloZup another way I was thinking was to move the extracted archive folder to `hana_extract_dir` directory using saltstack functionality. So moving `hana_extract_dir/51053381` to `hana_extract_dir` which would still need the jinja logic to find `51053381` foldername, but avoid putting this logics in all other states.
Let me know if you can think of better way.

arbulu89 (changes requested):
@Simranpal Create a macro for this duplicated code. 
Create a `macros` folder and inside `get_hana_extract_dir.sls` or similar, and we can use the macro. The macros basically substitute code, so it's exactly what you are doing here.

Inside, we can have something like:

```
{% macro get_hana_extract_dir(hana) -%}
{% set hana_extract_dir = hana.hana_extract_dir %}
# Below is temporary workaround to update the software installation path when using unrar for HANA multipart rar archive
# TODO: Find better solution to set or detect the correct extraction path when extracting multipart rar archive
# Below logic finds the extraction location based on name of multipart exe archive filename
{%- if hana.hana_archive_file is defined and hana.hana_archive_file.endswith((".exe", ".EXE")) %}
{% set archive_base_name = salt['file.basename']( hana.hana_archive_file.split('.')[0]) %}
{% set archive_name = archive_base_name.split('_')[0] %}
{% set hana_extract_dir = hana_extract_dir| path_join(archive_name) %}
{% endif %}
{%- endmacro %}
```

MalloZup (commented):
I'm was trying to understand the problem but I miss it a bit from the description.
so regarding your examplle

`51053381_part1.exe 51053381_part2.rar` if we extract them  to /tmp/foo and /tmp/bar cannot we move them afterward to `fobar` dir via salt mv? 
I didn't get the main issue here.

Also why upstream call of salt didn't work. if there is a bug there we can always comment with http link on code if we are working around ..

thx for clarification :pray:

Simranpal:
> I'm was trying to understand the problem but I miss it a bit from the description.
> so regarding your examplle
> 
> `51053381_part1.exe 51053381_part2.rar` if we extract them to /tmp/foo and /tmp/bar cannot we move them afterward to `fobar` dir via salt mv?
> I didn't get the main issue here.
> 
> Also why upstream call of salt didn't work. if there is a bug there we can always comment with http link on code if we are working around ..
> 
> thx for clarification


Multi part archives are extracted to 1 output folder. So in case of HANA installation multipart archive `51053381_part1.exe 51053381_part2.rar 51053381_part3.rar` with extract to `51053381` folder using the unrar tool.

The problem is that we cannot extract with unrar/ unar tool directly to` /sapmedia_extract/HANA`. It will create a top-level directory there like `/sapmedia_extract/HANA/51053381`.  If we can make the unrar tool to extract `/sapmedia_extract/HANA/ `instead of automatic folder `/sapmedia_extract/HANA/51053381` we don't need this workaround logic. Based on my tests with `unrar` and `unrar_wrapper`, the option does not work. And the salt upstream functionality does not support for multi-part rar archives. 

Since the hana installer needs this full path, and I am calculating the last part `51053381` with jinja logic. That logic was in hana extraction sls, but causing problems with install.sls in SUMA because some variables cannot be imported at highstate. To workaround this, I have created the variables inside the `hana/install.sls` and related states

Simranpal:
@arbulu89 I have updated to use the macros function, but I am not sure of location for the macros directory, and I have to update the spec file accordingly

MalloZup (approved):
looks good to me even the location.

MalloZup:
only there is a conflict on changelog

saphanabootstrap-formula.spec
```diff
@@ -36,6 +36,7 @@ Suggests:       prometheus-hanadb_exporter >= 0.7.0
 %define fname hana
 %define fdir  %{_datadir}/salt-formulas
 %define ftemplates templates
+%define fmacros macros
```
arbulu89, line 39: I don't think we need this change. All the content from `hana` folder should me already moved properly

saphanabootstrap-formula.spec
```diff
@@ -76,6 +77,7 @@ fi
 
 %attr(0755, root, salt) %{fdir}/states/%{fname}
 %attr(0755, root, salt) %{fdir}/states/%{fname}/%{ftemplates}
+%attr(0755, root, salt) %{fdir}/states/%{fname}/%{fmacros}
```
arbulu89, line 80: I think we don't need this change

arbulu89 (approved):
@Simranpal The implementation looks good. Just check if the changes in the `spec` file are needed. I don't think they are, the folder owner/permissions should be inherited.
Anyway, if you did the test with the 2 options, and they are mandatory, go ahead and merge, but please run the tests (if we can spare it, better).
The macros folder location is good in my opinion.

**Make sure you put the newest changes entry on the top, otherwise OBS will complain**

saphanabootstrap-formula.spec
Simranpal: thanks I will check it

Simranpal:
@arbulu89 the changes were not need in spec file with current macros directory location. Originally I was testing with macros folder in parent folder, and it was needed, however its not needed in current case and I removed it.

I have tested with `EXE`, `ZIP` SAP media extraction in `libvirt` and `azure`.

@MalloZup @arbulu89 for future enhancement, as users will request to provide others HANA components SAR files(like Hana client etc.. like HANA DB IMDB server SAR + HANA Client SAR file) we need to support that feature. I will make improvements then by moving all extraction logics to macros.

PR #109, opened 2020-10-08
Update the SUMA metadata to show hana form under SAP deployment group

Simranpal:
With this change, both SAP HANA and NW formulas will show under the SAP deployment group, under the SUMA formulas list

image


arbulu89 (approved):
Great! Thanks

PR #111, opened 2020-11-11
HANA client SAR archive extraction

Simranpal:
With respect to: https://github.com/SUSE/ha-sap-terraform-deployments/issues/558

I have added the option to install HANA Client packages from SAR file for monitoring, cost-optimized HANA.  This is needed if the HANA database is installed using a SAR file where the client is not present.
This option works when you are using HANA DB archive SAR file only. In other cases, HANA platform folder or package will already have the HANA client.

There are 3 options now to use HANA Client. Below is their preference order:
1) HANA platform folder/archives (except hana db sar archive)- HANA client already included with this. So no additional parameters required for this option
2) Extracted hana client folder provided via variable: `hana_client_software_path` 
3) HANA Client sar file provided via variable: `hana_client_archive_file`  

Some additional changes with this PR:
- Remove node specific hana_client option in the exporter entry from pillar
- Update pillar example description for hana media archives

Test with deployments in libvirt using the above options

### DIFF
--- a/_service
+++ b/_service
@@ -4,7 +4,7 @@
     <param name="scm">git</param>
     <param name="exclude">.git</param>
     <param name="filename">saphanabootstrap-formula</param>
-    <param name="versionformat">0.6.3+git.%ct.%h</param>
+    <param name="versionformat">0.6.2+git.%ct.%h</param>
     <param name="revision">%%VERSION%%</param>
   </service>
 
--- a/_service
+++ b/_service
@@ -4,7 +4,7 @@
     <param name="scm">git</param>
     <param name="exclude">.git</param>
     <param name="filename">saphanabootstrap-formula</param>
-    <param name="versionformat">0.6.2+git.%ct.%h</param>
+    <param name="versionformat">0.6.3+git.%ct.%h</param>
     <param name="revision">%%VERSION%%</param>
   </service>
 
--- a/hana/enable_cost_optimized.sls
+++ b/hana/enable_cost_optimized.sls
@@ -1,6 +1,10 @@
 {%- from "hana/map.jinja" import hana with context -%}
 {%- from 'hana/macros/get_hana_exe_extract_dir.sls' import get_hana_exe_extract_dir with context %}
-{% set hana_extract_dir = get_hana_exe_extract_dir(hana) %}
+{%- if hana.client_archive_file is defined and hana.hana_archive_file.endswith((".sar", ".SAR") %}
+{%- set hana_client_path = hana.client_extract_dir %}
+{%- else %}
+{%- set hana_client_path = get_hana_exe_extract_dir(hana) %}
+{%- endif %}
 
 {% set host = grains['host'] %}
 
@@ -56,7 +60,7 @@ failure:
 extract_hana_pydbapi_archive:
     hana.pydbapi_extracted:
       - name: PYDBAPI.TGZ
-      - software_folders: [{{ node.install.software_path|default(hana.software_path)|default(hana_extract_dir) }}]
+      - software_folders: [{{ node.install.software_path|default(hana.software_path)|default(hana_client_path) }}]
       - output_dir: /hana/shared/srHook
       - hana_version: '20'
       - force: true
--- a/hana/enable_cost_optimized.sls
+++ b/hana/enable_cost_optimized.sls
@@ -1,6 +1,6 @@
 {%- from "hana/map.jinja" import hana with context -%}
 {%- from 'hana/macros/get_hana_exe_extract_dir.sls' import get_hana_exe_extract_dir with context %}
-{%- if hana.client_archive_file is defined and hana.hana_archive_file.endswith((".sar", ".SAR")) %}
+{%- if hana.client_archive_file is defined and hana.hana_archive_file is defined and hana.hana_archive_file.endswith((".sar", ".SAR")) %}
 {%- set hana_client_path = hana.client_extract_dir %}
 {%- else %}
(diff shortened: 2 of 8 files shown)


Simranpal:
I have added an additional variable: `client_extract_dir`. I am thinking to use a subfolder under the existing `hana_archive_file` something like `/sapmedia_extract/HANA/CLIENT`. 
Let me know if you think if its better to ask user for HANA client extract directory? or just use the hana extract path

_service
```diff
@@ -4,7 +4,7 @@
     <param name="scm">git</param>
     <param name="exclude">.git</param>
     <param name="filename">saphanabootstrap-formula</param>
-    <param name="versionformat">0.6.2+git.%ct.%h</param>
+    <param name="versionformat">0.6.3+git.%ct.%h</param>
```
arbulu89, line 7: @Simranpal Let's try to avoid increasing the version here. It will be better we just increase when we are going to deliver and created a new submission.
We can group this way many changes in a new release

saphanabootstrap-formula.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Nov 11 04:25:03 UTC 2020 - Simranpal Singh <simranpal.singh@suse.com>
+
+- Version 0.6.3
```
arbulu89, line 4: The same, we will increase the version when we submit

pillar.example
```diff
@@ -18,6 +18,9 @@ hana:
   # hana_extract_dir should be a new directory and seperate from location where the compressed files are present, to avoid conflicts in file permissions.
   hana_archive_file: '/sapmedia/51053492.ZIP'
   hana_extract_dir: '/sapmedia_extract/HANA'
+  # The HANA Client sar archive is needed for monitoring if HANA DB sar archive is provided. It will be extracted to hana_client_extract_dir path (optional, by default /sapmedia_extract/HANA_CLIENT)
+  client_archive_file: '/sapmedia/IMDB_CLIENT20_003_144-80002090.SAR'
```
arbulu89, line 22: Maybe they should be `hana_client_archive_file` and `hana_client_extract_dir`?
In fact, the comment above uses that syntax.
To make it more consistent

hana/monitoring.sls
```diff
@@ -31,7 +35,7 @@ install_python_pip:
 extract_pydbapi_client:
   hana.pydbapi_extracted:
     - name: PYDBAPI.TGZ
-    - software_folders: [{{ exporter.hana_client_path|default(node.install.software_path)|default(hana.software_path)|default(hana_extract_dir) }}]
+    - software_folders: [{{ exporter.hana_client_path|default(node.install.software_path)|default(hana.software_path)|default(hana_client_path) }}]
```
arbulu89, line 38: I guess we will need to remove the `hana_client_path` from `exporter` entry, right?
They seem duplicated

hana/enable_cost_optimized.sls
```diff
@@ -1,6 +1,10 @@
 {%- from "hana/map.jinja" import hana with context -%}
 {%- from 'hana/macros/get_hana_exe_extract_dir.sls' import get_hana_exe_extract_dir with context %}
-{% set hana_extract_dir = get_hana_exe_extract_dir(hana) %}
+{%- if hana.client_archive_file is defined and hana.hana_archive_file is defined and hana.hana_archive_file.endswith((".sar", ".SAR")) %}
```
arbulu89, line 3: This condition looks pretty confusing. I would add maybe a comment above to say what it does.
I understand that if hana archive is sar format, and there is a hana client archive, it uses this path. Otherwise, it uses the hana platform path

arbulu89: Besides, I see this piece of code duplicated. Should we create a macro for this too?

arbulu89 (changes requested):
@Simranpal I think there are some things we need to sort out.
Now, we have a `hana_client_path` in the `exporter`. Definitely we need to remove this to use the new option.
Besides that, we must clarify somewhere that, if the HANA client is not present, the monitoring won't work with the HANA DB sar option.

Finally, what if the user provides an already extracted HANA client? Did we consider this option. Otherwise, we must clarify that the HANA client must come in sar format

hana/monitoring.sls
Simranpal: We can leave it to provide node specific `hana_client_path` right

hana/enable_cost_optimized.sls
Simranpal: I am not sure about that. This code is calling the previous macro. Nested macros maybe confusing

arbulu89: Why? I don't see them confusing. It's quite natural in programming (methods calling other methods).
Actually having duplicated code in pieces that we don't look often looks risky. Most probably we would forget to do the changes

arbulu89 (approved):
Hi @Simranpal ,
I still think we should encapsulate the duplicated code in a macro, it seems natural to me.
About the `hana_client` in the `exporter` entry, I would remove it, but this up to you.

Simranpal:
thanks @arbulu89 I have created the macros and added support for already extracted HANA client. I have updated the pillar example help and the PR description. Let me know how it looks

hana/enable_cost_optimized.sls
Simranpal: yes I tested the code with macro. And from maintenance purpose as well it makes sense to add the macro for this code

arbulu89 (approved):
Great @Simranpal !

PR #113, opened 2020-12-02
Start the saptune systemd service after saptune solution is applied

Simranpal:
Currently, the deployment doesn't start the saptune daemon. It is needed to get the sap tuning back especially after system reboot.
I have added the `saptune daemon start` cmd. which will be available in saptune version 3, but internally mapped to the new command `saptune service start`

Resolves the `ERROR: The parameters listed above have deviated from SAP/SUSE recommendations.`

### DIFF
--- a/hana/saptune.sls
+++ b/hana/saptune.sls
@@ -11,4 +11,8 @@ apply_saptune_solution_{{ host }}_{{ name }}:
   saptune.solution_applied:
     - name: {{ saptune_solution }}
 
+start_saptune_daemon_{{ host }}_{{ name }}:
+  cmd.run:
+    - name: saptune daemon start
+
 {% endfor %}
--- a/hana/saptune.sls
+++ b/hana/saptune.sls
@@ -11,7 +11,8 @@ apply_saptune_solution_{{ host }}_{{ name }}:
   saptune.solution_applied:
     - name: {{ saptune_solution }}
 
-start_saptune_daemon_{{ host }}_{{ name }}:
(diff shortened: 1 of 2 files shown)


hana/saptune.sls
```diff
@@ -11,4 +11,8 @@ apply_saptune_solution_{{ host }}_{{ name }}:
   saptune.solution_applied:
     - name: {{ saptune_solution }}
 
+start_saptune_daemon_{{ host }}_{{ name }}:
```
MalloZup, line 14: I would like maybe here that we add a comment/rationale why we are starting the daemon.

E.g I would add here:

```
# start saptune daemon will ensure that after a reboot of the host, the solution applied via saptune is still compliant

```

MalloZup (changes requested):
I would like to see a comment explaining why we add it

MalloZup (commented):
a part of that looks good to me

hana/saptune.sls
angelabriel, line 14: even that the command option (for now) is called 'daemon', I would prefer to change the label from 'start_saptune_daemon_' to 'start_saptune_service_' as this reflects much better, what the command really does. It starts the systemd service  'tuned.service' (in saptune version 2) and in the future it starts the systemd service 'saptune.service' (saptune version 3)

angelabriel (commented):
beside my comment above, looks good to me. Thanks

MalloZup (approved):
thx!

arbulu89 (approved):
Thanks @Simranpal

PR #117, opened 2021-01-19
Replace require by recommend for salt-formulas-configuration for SLE12

arbulu89:
Replace Require by Recommend for salt-formulas-configuration for SLE12.

https://bugzilla.suse.com/show_bug.cgi?id=1177860


Simranpal (approved):
Thanks! looks good!

PR #120, opened 2021-03-05
Fix hana client preference usage to use first the defined client method

arbulu89:
In some scenarios the hana client is not found properly. 
- HANA client and server SAR archives are already extracted in the sapmedia folder
- HANA database SAR is extracted and HANA client is in SAR format

To fix that I have done some preference changes. Now, if the `hana_client_archive_file` or `hana_client_software_path
` are present they will have preference over the database/platform options.

The change includes:
1. The HANA client is always extracted if the data is present. Before it was extracted only if the database was a SAR option
2. `hana_client_archive_file` or `hana_client_software_path
` have preference over other options

I have tested the next scenarios and everything worked fine:

Database as SAR - Client as SAR -> OK
Database as SAR - Client already extracted -> OK
Database already extracted - Client as SAR -> OK (before this was failing)
Database already extracted - Client already extracted -> OK (before this was failing)

The database can be in simple version (the patch option) or as platform, both work the same way.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -10,23 +10,25 @@ hana:
   saptune_solution: 'HANA'
 
   # HANA installation media can be provided in one of the two methods: extracted HANA platform folder or HANA media archive
-  # 1. Path to already extracted HANA platform installation media. This will have preference over hana installation media archive
+  # 1. Path to an already extracted HANA platform installation media. This will have preference over hana installation media archive
   software_path: '/sapmedia/HANA/51052481'
   # 2. Or specify the path to the hana installation media archive
   # If using hana sar archive, please also provide compatible version of sapcar executable
   # The archive will be extracted to path specified at hana_extract_dir (optional, by default /sapmedia_extract/HANA)
-  # hana_extract_dir should be a new directory and seperate from location where the compressed files are present, to avoid conflicts in file permissions.
+  # hana_extract_dir should be a new directory and separated from the location where the compressed files are present, to avoid conflicts in file permissions.
   hana_archive_file: '/sapmedia/51053492.ZIP'
   hana_extract_dir: '/sapmedia_extract/HANA'
-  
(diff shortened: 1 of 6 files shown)


stefanotorresi (approved):
LGTM!

arbulu89:
> The changes look good. Below is two comments:
> 
>     1. We need to update `pillar.example` with comment about the new preference
> 
>     2. With this we bring one change: The hana client version maybe different from existing Hana platform provided client but I think this is users' responsibility to provide the compatible version of version. Adding the comment from point 1. can inform them.

@Simranpal Pillar updated in https://github.com/SUSE/saphanabootstrap-formula/pull/120/commits/fb4ffd92064cd5ed4fccd09ac72c44028ce1c02c
Could you have a new look and see?

pillar.example
```diff
@@ -10,23 +10,25 @@ hana:
...
+  # hana_extract_dir should be a new directory and separated from the location where the compressed files are present, to avoid conflicts in file permissions.
   hana_archive_file: '/sapmedia/51053492.ZIP'
   hana_extract_dir: '/sapmedia_extract/HANA'
-  
-  # HANA Client packages are needed for monitoring & cost-optimized scenario. HANA Client is already included in HANA platform media unless a HANA database sar archive is used 
+
+  # HANA Client packages are needed for monitoring & cost-optimized scenario. HANA Client is already included in HANA platform media unless a HANA database sar archive is used
   # If the HANA archive used for installation is sar format specified above, you need to provide HANA Client in one of the two ways: extracted HANA client folder or HANA client sar archive file
+  # If any of the next two options are used, this HANA Client will have preference over the HANA Client coming in the HANA platform described above. If HANA platform is used, it is usually better to
+  # not using the HANA Client as it might bring some compatibility issue.
```
Simranpal, line 25: better to `not using / not use`

Simranpal (approved):
Looks good, just minor typo

PR #121, opened 2021-03-11
Feature/gcp load balancer

arbulu89:
Implement the load balancer use case for GCP. Some info at: https://cloud.google.com/solutions/sap/docs/sap-hana-ha-config-sles

I have done some refactoring of the template file. It was quite difficult to keep adding things, so I have splited the resources depending on which cloud provider is used. It creates some duplicated code, but I still think that it is easier to understand and continue adding things. 
This is done in: https://github.com/SUSE/saphanabootstrap-formula/commit/c047b1a84b9db49d3aa108d84d23d219a45bb25a

The load balancer implementation comes in: https://github.com/SUSE/saphanabootstrap-formula/commit/07cbdd1cb109d1de4e3b62a022713e98dce5bf99

Implemented for: https://github.com/SUSE/ha-sap-terraform-deployments/issues/625

**Disclaimer: I know that I have not added any documentation about how to use it. I think we need to discuss this outside of this PR. The whole templating thing for the cluster must be re-designed in my opinion to improve the documentation and usability**

### DIFF
--- a/templates/scale_up_resources.j2
+++ b/templates/scale_up_resources.j2
@@ -5,9 +5,9 @@
 {%- set native_fencing = data.native_fencing|default(True) %}
 {%- set monitoring_enabled = pillar.cluster.monitoring_enabled|default(False) %}
 
-#
-# defaults
-#
+###########
+# Defaults
+###########
 
 rsc_defaults \
     resource-stickiness="1000" \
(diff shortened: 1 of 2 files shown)


templates/scale_up_resources.j2
```diff
@@ -69,8 +94,15 @@ primitive rsc_gcp_vip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:gcp-vpc-move-rou
 
 colocation col_saphana_ip_{{ sid }}_HDB{{ instance }} 2000: rsc_gcp_vip_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Master
 
+{%- endif %}
```
Simranpal, line 97: I would add a new line or maybe 2 lines, after this, for better readability

templates/scale_up_resources.j2
```diff
@@ -160,17 +197,45 @@ primitive rsc_ip_{{ sid }}_HDB{{ instance }}_readenabled ocf:suse:aws-vpc-move-i
...
     params ip={{ data.virtual_ip_secondary }} vpc_network={{ data.vpc_network_name }} route_name={{ data.route_name_secondary }} \
     op start interval=0 timeout=180 \
     op stop interval=0 timeout=180 \
     op monitor interval=60 timeout=60
 
+colocation col_saphana_ip_{{ sid }}_HDB{{ instance }}_readenabled 2000: \
+    rsc_ip_{{ sid }}_HDB{{ instance }}_readenabled:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Slave
+
 {%- endif %}
+{%- elif cloud_provider == "microsoft-azure" %}
```
Simranpal, line 234: I would put new line here too

Simranpal (approved):
The structure looks okay to me. Better than before.
Regarding the new resource agents values, I could not find all the values for GCP which you used. Are there another references? or based on tests

PR #122, opened 2021-03-12
Fix how sidadm user is transformed to lowercase in ha cluster states

arbulu89:
Related to: https://github.com/SUSE/shaptools/pull/65

Fix some sidadm conversions. If the provided sid is in uppercase the created states are incorrect, specifically the ones managing the `sudoers` entry content.

The rest of the uses of the sidadm are safe as they are treated in lower layers (`shaptools` specifically)

Some info: https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-CostOpt-12/#_allowing_sidadm_to_access_the_cluster

FYI: @petersatsuse

### DIFF
--- a/hana/ha_cluster.sls
+++ b/hana/ha_cluster.sls
@@ -53,7 +53,7 @@ sudoers_backup_{{ sap_instance }}:
   file.copy:
     - name: {{ tmp_sudoers }}
     - source: {{ sudoers }}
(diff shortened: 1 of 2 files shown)


Simranpal (approved):
Thanks, yes I see our salt `file` managing states needed to be updated. Rest cases are handled at lower level

PR #124, opened 2021-04-28
Set correct stickiness for the azure-lb resource

diegoakechi:
The azure-lb resource receives an stickiness=0 to not influence on
transitions calculations as the HANA resources have more priority.
Ref: https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources

### DIFF
--- a/saphanabootstrap-formula.changes
+++ b/saphanabootstrap-formula.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr 28 14:26:18 UTC 2021 - Diego Akechi <dakechi@suse.com>
+
+- Set correct stickiness for the azure-lb resource
+  The azure-lb resource receives an stickiness=0 to not influence on
+  transitions calculations as the HANA resources have more priority
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


arbulu89 (changes requested):
Hi @diegoakechi ,
Please, add the applied change in the changelog first.
Besides that, I guess this applies to NW clusters too, right?

**Edit**: I think we have the same problem with the hanadb_exporter. We could do the same I guess (the error is tracked in our internal jira)

diegoakechi:
@arbulu89 For NW the stickness calculation is different and there is no recommendation for the Azure-lb in that regard. See: https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse

About the https://github.com/SUSE/ha-sap-terraform-deployments/issues/686 stickiness=0 solves part of the problem, but it will also need a change on the collocation constraint, which will demand more testing, so I think a specific PR would be the best here. AFAIR @pirat013 will work on it. 

I will include the changelog change.

arbulu89 (approved):
Thanks

PR #125, opened 2021-04-30
prometheus-hanadb_exporter cluster primitiv adoption fix SUSE/ha-sap-terraform-deployments/#/686

pirat013:
prometheus-hanadb_exporter cluster primitiv adoption, the fix will address this issue:
https://github.com/SUSE/ha-sap-terraform-deployments/issues/686
The change is needed to avoid a HANA DB takeover just in case that the local hanadb_exporter has an issue and is not able to start.

### DIFF
--- a/hana/monitoring.sls
+++ b/hana/monitoring.sls
@@ -62,7 +62,7 @@ prometheus-hanadb_exporter:
 hanadb_exporter_configuration_{{ exporter_instance }}:
   file.managed:
     - source: salt://hana/templates/hanadb_exporter.j2
-    - name: /etc/hanadb_exporter/{{ exporter_instance }}.json
+    - name: /usr/etc/hanadb_exporter/{{ exporter_instance }}.json
     - template: jinja
     - require:
       - prometheus-hanadb_exporter
--- a/hana/monitoring.sls
+++ b/hana/monitoring.sls
@@ -62,7 +62,7 @@ prometheus-hanadb_exporter:
 hanadb_exporter_configuration_{{ exporter_instance }}:
   file.managed:
     - source: salt://hana/templates/hanadb_exporter.j2
-    - name: /usr/etc/hanadb_exporter/{{ exporter_instance }}.json
+    - name: /etc/hanadb_exporter/{{ exporter_instance }}.json
     - template: jinja
     - require:
       - prometheus-hanadb_exporter
--- a/templates/scale_up_resources.j2
+++ b/templates/scale_up_resources.j2
@@ -270,8 +270,9 @@ primitive rsc_exporter_{{ sid }}_HDB{{ instance }} systemd:prometheus-hanadb_exp
     op start interval=0 timeout=100 \
     op stop interval=0 timeout=100 \
     op monitor interval=10 \
+    meta resource-stickiness=0 \
     meta target-role=Started
 
-colocation col_exporter_{{ sid }}_HDB{{ instance }} +inf: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Master
+colocation col_exporter_{{ sid }}_HDB{{ instance }} 1: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Master
 
 {%- endif %}
--- a/templates/scale_up_resources.j2
+++ b/templates/scale_up_resources.j2
@@ -273,6 +273,6 @@ primitive rsc_exporter_{{ sid }}_HDB{{ instance }} systemd:prometheus-hanadb_exp
     meta resource-stickiness=0 \
     meta target-role=Started
 
-colocation col_exporter_{{ sid }}_HDB{{ instance }} 1: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Master
+colocation col_exporter_{{ sid }}_HDB{{ instance }} -inf: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Slave
 
 {%- endif %}
--- a/saphanabootstrap-formula.changes
+++ b/saphanabootstrap-formula.changes
@@ -4,7 +4,6 @@ Fri May 07 07:15:21 UTC 2021 - Bernd Schubert <bschubert@suse.com>
 - Changing exporter stickiness to => 0 and adjusting the colocation 
   score from +inf to -inf and changing the colocation from Master to Slave. 
   This change fix the impact of a failed exporter in regards to the HANA DB.
-- changing path for hanadb_exporter_configuration as described in the documentation
 -------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


arbulu89 (commented):
Hi @pirat013 

I have a proposal here. As the `hanadb_exporter` resource must always run alongside the SAP master and we have the problem that the resource is moving to the 2nd node, we can add an additional `colocation` to avoid that, saying that this resource cannot go together with the `Slave`.

Something like:
```
colocation col_exporter_{{ sid }}_HDB{{ instance }}_Slave -inf: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Slave
```

This goes together with your changes. This way, the resource will stop working instead of moving to node2, what we don't want.

**Side notes**
- A failure in the exporter is really rare. Incorrect, host, port or password don't make the service fail, they just continue retrying forever but the resource state would be green. Just to raise some awareness
- We would need a bgz ticket ID to publish this change on SLE

pirat013:
Hi @arbulu89,

I trust you when you say that the exporter  fails is rare. That's not the only point here. What we should avoid is that an "unimportant" service could influence the HANA DB start / stop or takeover. 
I'm not sure if the colocation constraint for the secondary (HANA slave) does not have such an impact as it is the case with the current colocation rule for the HANA Master.
If we would like to have the exporter controlled by Pacemaker and let them run somewhere in the cluster independently we just have to configure the exporter pointing to the vIP of HANA and configure two possible exporters in the Prometheus config. In that case we would see all the time one Prometheus_exporter who is configured, is not running. But than we are absolutely independent from the HANA SR resource and the exporter does not has any impact for the HANA takeover score calculation and so one.

arbulu89:
@pirat013 Yes, I get your point and purpose of this PR is clear to me
- An error in the exporter shouldn't impact the DB
- The exporter must be always running in the same host than the Master resource

With my proposal this is guaranteed (I think it works even without the `Master` colocation) and the resource won't have any impact as the stickiness is 0.

If your proposal is to not have the exporter managed by the cluster this PR doesn't make much sense, and we should remove the resource completely. If we have both configured and working always, and the prometheus server scraping in both, why do we need the exporter in the cluster?
In any case, this would be a worst usage in general, as the HANA grafana dashboard depends on having a fixed exporter data to visualize properly the database information.

Finally, we shouldn't be scared of manipulating the Cluster resources as long as we do it fine. We are experts in that area.

diegoakechi:
@pirat013 @arbulu89  Nice and trick discussion here =). I like the idea of changing the perspective to avoid running the exporter on the secondary node. This way we are preserving HANA resource but also avoiding that the exporter migrates to other nodes, knowing that it will not work. I vote for this path.

pirat013:
I guess we have an agreement :-) Thanks @arbulu89 lets take your proposal.

arbulu89:
Hey @pirat013 ,
Could you apply the proposed changes in this PR so we can move forward the implementation?
Thanks

hana/monitoring.sls
```diff
@@ -62,7 +62,7 @@ prometheus-hanadb_exporter:
 hanadb_exporter_configuration_{{ exporter_instance }}:
   file.managed:
     - source: salt://hana/templates/hanadb_exporter.j2
-    - name: /usr/etc/hanadb_exporter/{{ exporter_instance }}.json
+    - name: /etc/hanadb_exporter/{{ exporter_instance }}.json
```
arbulu89, line 65: We cannot change this

templates/scale_up_resources.j2
```diff
@@ -273,6 +273,6 @@ primitive rsc_exporter_{{ sid }}_HDB{{ instance }} systemd:prometheus-hanadb_exp
     meta resource-stickiness=0 \
     meta target-role=Started
 
-colocation col_exporter_{{ sid }}_HDB{{ instance }} 1: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Master
+colocation col_exporter_{{ sid }}_HDB{{ instance }} -inf: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Slave
```
arbulu89, line 276: I think we didn't understand each other. What I suggest is the next:

```
primitive rsc_exporter_{{ sid }}_HDB{{ instance }} systemd:prometheus-hanadb_exporter@{{ sid }}_HDB{{ instance }} \
    op start interval=0 timeout=100 \
    op stop interval=0 timeout=100 \
    op monitor interval=10 \
    meta resource-stickiness=0 \
    meta target-role=Started

colocation col_exporter_{{ sid }}_HDB{{ instance }} 1: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Master

colocation col_exporter_{{ sid }}_HDB{{ instance }}_Slave -inf: rsc_exporter_{{ sid }}_HDB{{ instance }}:Started msl_SAPHana_{{ sid }}_HDB{{ instance }}:Slave

```

Basically, I proposed to add an additional `colocation` to the ones you already added keeping the `resource-stickines=0` value too.
At least this is what I tested and worked.

arbulu89 (changes requested):

templates/scale_up_resources.j2
arbulu89: If you tried only with the `Slave` colocation and it works it is fine for anyway.

hana/monitoring.sls
pirat013: Okay, but it does not match to the documentation and the same is true for the logging-configuration.

templates/scale_up_resources.j2
pirat013: I got your point. But this would add more complexity to the cluster and dependencies between HANA master and slave. I've tested the proposed change and have added only the colocation constraint for the slave and it works.

hana/monitoring.sls
arbulu89: It is documented here: https://github.com/SUSE/hanadb_exporter#running-as-a-daemon

Better to keep the config files under `/usr/etc` to avoid issues with package updates that bring the configuration file

templates/scale_up_resources.j2
arbulu89: Ok

arbulu89 (approved):
@pirat013 Thanks! We can merge after the conflict in the `changes` file is resolved.
Just keep the chronological order of the changes entry fine (newest at the top)

arbulu89 (changes requested):
@pirat013 The changes file is not correctly formatted.

The content should be:
```
-------------------------------------------------------------------
Fri May 07 09:15:21 UTC 2021 - Bernd Schubert <bschubert@suse.com>

- Changing exporter stickiness to => 0 and adjusting the colocation 
  score from +inf to -inf and changing the colocation from Master to Slave. 
  This change fix the impact of a failed exporter in regards to the HANA DB.

-------------------------------------------------------------------
Thu May  6 08:49:19 UTC 2021 - Xabier Arbulu <xarbulu@suse.com>

- Change hanadb_exporter default timeout value to 30 seconds

-------------------------------------------------------------------
```

arbulu89 (approved):

diegoakechi:
Thanks @pirat013 !

PR #126, opened 2021-05-06
Change hanadb_exporter start timeout value to 30s

arbulu89:
Simple change to reduce the hanadb_exporter start timeout to 30 seconds
(And it took advantage of fixing one of the states involved in the exporter initialization)


stefanotorresi (approved):
lgtm

PR #130, opened 2021-06-18
add native fencing for microsoft-azure

yeoldegrove:
This adds the needed resources to implement native fencing on azure.

It also defines for all cloud providers the same defaults to use native fencing as https://github.com/SUSE/ha-sap-terraform-deployments/.
```
# ha-sap-terraform-deployments
grep -A5 hana_cluster_fencing_mechanism */variables.tf | grep 'variable |default'
  1:openstack/variables.tf:variable "hana_cluster_fencing_mechanism" {
  4:openstack/variables.tf-  default     = "sbd"
  8:libvirt/variables.tf:variable "hana_cluster_fencing_mechanism" {
  11:libvirt/variables.tf-  default     = "sbd"
  21:gcp/variables.tf:variable "hana_cluster_fencing_mechanism" {
  24:gcp/variables.tf-  default     = "native"
  34:azure/variables.tf:variable "hana_cluster_fencing_mechanism" {
  37:azure/variables.tf-  default     = "sbd"
  47:aws/variables.tf:variable "hana_cluster_fencing_mechanism" {
  50:aws/variables.tf-  default     = "native"
```

```
# this formula
{%- if cloud_provider == "amazon-web-services" %}
{%- set native_fencing = data.native_fencing|default(True) %}
{%- elif cloud_provider == "google-cloud-platform" %}
{%- set native_fencing = data.native_fencing|default(True) %}
{%- elif cloud_provider == "microsoft-azure" %}
{%- set native_fencing = data.native_fencing|default(False) %}
{%- endif %}

```


diegoakechi (commented):
@yeoldegrove @arbulu89 only read it quickly, but looks we are removing SBD support on Azure?  SBD is still supported on Azure and preferred by some customers, so I would say we need to have some default for sure, but giving the option to choose.

yeoldegrove:
The idea is to have "SBD" as the default value for azure. Native fencing should only be an option.
I'll double check the code to make sure this is the case.

diegoakechi:
Thanks... Just checking the understanding. Didn't read all the changes.

yeoldegrove:
One could add something like this to make sure there are the defaults that we want:
```
{%- if cloud_provider == "amazon-web-services" %}
{%- set native_fencing = data.native_fencing|default(True) %}
{%- elif cloud_provider == "google-cloud-platform" %}
{%- set native_fencing = data.native_fencing|default(True) %}
{%- elif cloud_provider == "microsoft-azure" %}
{%- set native_fencing = data.native_fencing|default(False) %}
{%- endif %}
```

@stefanotorresi @arbulu89 What do you think?

arbulu89:
> One could add something like this to make sure there are the defaults that we want:
> 
> ```
> {%- if cloud_provider == "amazon-web-services" %}
> {%- set native_fencing = data.native_fencing|default(True) %}
> {%- elif cloud_provider == "google-cloud-platform" %}
> {%- set native_fencing = data.native_fencing|default(True) %}
> {%- elif cloud_provider == "microsoft-azure" %}
> {%- set native_fencing = data.native_fencing|default(False) %}
> {%- endif %}
> ```
> 
> @stefanotorresi @arbulu89 What do you think?

Hi @yeoldegrove ,
Yes, we could do something like this. About `native` or `sbd` being the default option, I cannot really judge. This should come from Azure or from our cloud architect I guess. 
Other question is if the fence agents should be mutual exclusive, or the 2 types could coexist.
What do you think @peteratsuse ?

Besides that,
The new azure fence agent must be used based on this new parameter, and not always.

arbulu89 (changes requested):
Hi @yeoldegrove ,
Nice!
A version bump must come with the version string update in the `_service` file.
https://github.com/SUSE/saphanabootstrap-formula/blob/master/_service#L7

yeoldegrove:
> A version bump must come with the version string update in the `_service` file.
> https://github.com/SUSE/saphanabootstrap-formula/blob/master/_service#L7

done

arbulu89 (approved):
Thanks @yeoldegrove 
I think that we are missing some documentation about the usage of these new paremeters, that maybe should go in the `habootstrap-formula` project.
But we will fix that in the future

arbulu89:
Hi @yeoldegrove ,
As commented in https://github.com/SUSE/sapnwbootstrap-formula/pull/87, we will need to do some changes here. Open a new PR with the updates

PR #131, opened 2021-06-29
add native fencing for microsoft-azure - follow up

yeoldegrove:
This is a follow up for #130 which uses pillars set in `habootstrap-formula` instead of grains.

This adds the needed resources to implement native fencing on azure.

It also defines for all cloud providers the same defaults to use native fencing as https://github.com/SUSE/ha-sap-terraform-deployments/.
```
# ha-sap-terraform-deployments
grep -A5 hana_cluster_fencing_mechanism */variables.tf | grep 'variable |default'
  1:openstack/variables.tf:variable "hana_cluster_fencing_mechanism" {
  4:openstack/variables.tf-  default     = "sbd"
  8:libvirt/variables.tf:variable "hana_cluster_fencing_mechanism" {
  11:libvirt/variables.tf-  default     = "sbd"
  21:gcp/variables.tf:variable "hana_cluster_fencing_mechanism" {
  24:gcp/variables.tf-  default     = "native"
  34:azure/variables.tf:variable "hana_cluster_fencing_mechanism" {
  37:azure/variables.tf-  default     = "sbd"
  47:aws/variables.tf:variable "hana_cluster_fencing_mechanism" {
  50:aws/variables.tf-  default     = "native"
```

```
# this formula
{%- if cloud_provider == "amazon-web-services" %}
{%- set native_fencing = data.native_fencing|default(True) %}
{%- elif cloud_provider == "google-cloud-platform" %}
{%- set native_fencing = data.native_fencing|default(True) %}
{%- elif cloud_provider == "microsoft-azure" %}
{%- set native_fencing = data.native_fencing|default(False) %}
{%- endif %}

```

PR #132, opened 2021-06-30
add native fencing for other providers like openstack

melzer-b1:
Deployment failed due unset native_fencing variable if openstack is used.


yeoldegrove:
Thanks @melzer-b1 !
I will cherry-pick this for #131 as it includes other fixes.

PR #133, opened 2021-07-30
HANA scale-out

yeoldegrove:
This is work in progress and might need additional changes.

### DIFF
--- a/hana/ha_cluster.sls
+++ b/hana/ha_cluster.sls
@@ -1,6 +1,21 @@
 {%- from "hana/map.jinja" import hana with context -%}
 {% set host = grains['host'] %}
 
+{% if hana.scale_out %}
+{% set sr_hook_path = '/usr/share/SAPHanaSR-ScaleOut' %}
+remove_SAPHanaSR:
+  pkg.removed:
+    - name: SAPHanaSR
+
+install_SAPHanaSR_ScaleOut:
+  pkg.installed:
+    - pkgs:
+      - SAPHanaSR-ScaleOut
+      - SAPHanaSR-ScaleOut-doc
+{% else %}
+{% set sr_hook_path = '/usr/share/SAPHanaSR' %}
+{% endif %}
+
 {% for node in hana.nodes if node.host == host %}
 
 {% set instance = '{:0>2}'.format(node.instance) %}
@@ -25,7 +40,7 @@ configure_ha_hook_{{ sap_instance }}:
 
         [ha_dr_provider_SAPHanaSR]
         provider = SAPHanaSR
-        path = /usr/share/SAPHanaSR
+        path = {{ sr_hook_path }}
         execution_order = 1
 
         [trace]
--- a/hana/ha_cluster.sls
+++ b/hana/ha_cluster.sls
@@ -78,7 +78,15 @@ sudoers_append_{{ sap_instance }}:
   file.append:
     - name: {{ tmp_sudoers }}
     - text: |
+        {% if hana.scale_out %}
+        # SAPHanaSR-ScaleOut needs for srHook
+        Cmnd_Alias SOK   = /usr/sbin/crm_attribute -n hana_{{ node.sid.lower() }}_glob_srHook -v SOK   -t crm_config -s SAPHanaSR
+        Cmnd_Alias SFAIL = /usr/sbin/crm_attribute -n hana_{{ node.sid.lower() }}_glob_srHook -v SFAIL -t crm_config -s SAPHanaSR
+        {{ node.sid.lower() }}adm ALL=(ALL) NOPASSWD: SOK, SFAIL
+        {% else %}
+        # SAPHanaSR needs for srHook
         {{ node.sid.lower() }}adm ALL=(ALL) NOPASSWD: /usr/sbin/crm_attribute -n hana_{{ node.sid.lower() }}_site_srHook_*
+        {% endif %}
     - require:
       - sudoers_backup_{{ sap_instance }}
 
--- a/hana/ha_cluster.sls
+++ b/hana/ha_cluster.sls
@@ -5,7 +5,9 @@
 {% set sr_hook_path = '/usr/share/SAPHanaSR-ScaleOut' %}
 remove_SAPHanaSR:
   pkg.removed:
-    - name: SAPHanaSR
+    - pkgs:
+      - SAPHanaSR
+      - SAPHanaSR-doc
 
 install_SAPHanaSR_ScaleOut:
   pkg.installed:
--- a/templates/cluster_resources.j2
+++ b/templates/cluster_resources.j2
@@ -4,6 +4,14 @@
 {%- set cloud_provider = grains['cloud_provider'] %}
 {%- set monitoring_enabled = pillar.cluster.monitoring_enabled|default(False) %}
 
+{%- set scale_out = data.scale_out|default(False)%}
+{%- set majority_maker = data.majority_maker|default("")%}
+{%- if scale_out %}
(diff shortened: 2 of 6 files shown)


hana/ha_cluster.sls
```diff
@@ -1,6 +1,23 @@
...
+    - pkgs:
+      - SAPHanaSR
+      - SAPHanaSR-doc
+
+install_SAPHanaSR_ScaleOut:
+  pkg.installed:
+    - pkgs:
+      - SAPHanaSR-ScaleOut
+      - SAPHanaSR-ScaleOut-doc
+{% else %}
```
arbulu89, line 17: Let's do one thing. As we started installing the packages for scale out, let's install as well for scale up (the `SAPHanaSR`). This is done now in the `habootstrap-formula`, but to continue with the same consistency, let's install them here. It makes sense

arbulu89 (commented):
Hi @yeoldegrove ,
Some minor comments.
In general it looks good

hana/ha_cluster.sls
```diff
@@ -61,7 +78,15 @@ sudoers_append_{{ sap_instance }}:
   file.append:
     - name: {{ tmp_sudoers }}
     - text: |
+        {% if hana.scale_out %}
```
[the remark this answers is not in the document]
yeoldegrove: :+1:

hana/ha_cluster.sls
yeoldegrove: Resource-Agents are removed from the pillars for `habootstrap-formula` and installed here now.

templates/cluster_resources.j2
```diff
@@ -168,10 +180,14 @@ primitive rsc_SAPHana_{{ sid }}_HDB{{ instance }} ocf:suse:SAPHana \
     op monitor interval="60" role="Master" timeout="700" \
     op monitor interval="61" role="Slave" timeout="700"
 
-ms msl_SAPHana_{{ sid }}_HDB{{ instance }} rsc_SAPHana_{{ sid }}_HDB{{ instance }} \
+ms msl_{{ SAPHanaResource }}_{{ sid }}_HDB{{ instance }} rsc_{{ SAPHanaResource }}_{{ sid }}_HDB{{ instance }} \
+{%- if scale_out %}
+    meta master-max="1" clone-node-max="1" interleave="true"
```
nick-wang, line 185: `master-max` is alias for `promoted-max` and for backward compatibility, maybe/will be deprecated in future. So if the `scale_out` is not target for old version, maybe use `promoted-max` instead is better. 

Refer to https://www.clusterlabs.org/pacemaker/doc/deprecated/en-US/Pacemaker/2.0/html/Pacemaker_Explained/_clone_options.html for more details.

yeoldegrove: This is based on https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-scaleOut-PerfOpt-12/index.html.
The old discussion... What shall be changed first? The docs? The code?

nick-wang: Well, the doc still title as SLE12 SP2 or newer ... Even it has already end of lifecycle (https://www.suse.com/lifecycle)

So we can keep `master-max` here to align with docs.

hana/ha_cluster.sls
```diff
@@ -78,7 +98,18 @@ sudoers_edit_{{ sap_instance }}:
     - force: true
     - require:
       - sudoers_check_{{ sap_instance }}
-      - stop_hana_{{ sap_instance }}
+
+# Stop SAP Hana
```
arbulu89, line 102: Did you check is this posterior `stop` works as expected and doesn't affect at any point the HANA processes?
In the documentation the stop is done before all of these updates (and this is how it was implemented).

If doing it later works, I'm fine with it, but I would put a more explicit comment here. Like `# Stop the HANA after applying the Cluster hooks. In the official guide this is done before the changes, but in order to make things easier, it is moved at the end`

arbulu89 (approved):
@yeoldegrove It looks good to me!

hana/ha_cluster.sls
yeoldegrove: @arbulu89 Both parameters that we set (system_replication.operation_mode and ha_dr_provider) need a restart of the HANA if touched (https://help.sap.com/viewer/009e68bc5f3c440cb31823a3ec4bb95b/2.0.05/en-US/514ab38a2e574c85a70ebba80ff16d99.html).
I tried it out several times and also know from my past experience, that changing the global.ini is safe on a running system. You just need the restart to activate certain changes.
A conditional restart is of course only possible at the very end after changes are applied (or not).

yeoldegrove: I also changed the way the `global.ini` file is processed to be idempotent, which makes me confident that this is an improvement.

yeoldegrove:
@arbulu89 Please recheck if we can merge this like this or if I shall split up this PR.

hana/ha_cluster.sls
arbulu89: Ok, nice! Thank you for this nice explanation

hana/enable_cost_optimized.sls
```diff
@@ -54,18 +54,19 @@ failure:
...
-
-          [ha_dr_provider_srCostOptMemConfig]
-          provider = srCostOptMemConfig
-          path = /hana/shared/srHook
-          execution_order = 2
-      - require:
-        - reduce_memory_resources_{{ node.host+node.sid }}
-        - setup_srHook_directory
-        - install_srCostOptMemConfig_hook
+  ini.options_present:
```
arbulu89, line 57: Nice! I was not aware of this `ini.options_present` state

hana/monitoring.sls
```diff
@@ -1,34 +1,32 @@
 {%- from "hana/map.jinja" import hana with context -%}
 {%- from 'hana/macros/get_hana_client_path.sls' import get_hana_client_path with context %}
 
-{% set pydbapi_output_dir = '/tmp/pydbapi' %}
-
-{% for node in hana.nodes if node.host == grains['host'] %}
-{%- set hana_client_path = get_hana_client_path(hana, node) %}
+{%- set pydbapi_output_dir = '/tmp/pydbapi' %}
+# first element in nodes has to be the primary
+{%- set node0 = hana.nodes[0] %}
```
arbulu89, line 6: This looks a potential source of errors.
We don't know if we will always have 2 nodes in the pillar file. Or even if we will have more nodes (potentially scale out, as this PR is working on)

hana/packages.sls
```diff
@@ -14,6 +14,8 @@ patterns-sap-hana:
     - retry:
         attempts: 3
         interval: 15
+    # SAPHanaSR-ScaleOut conflicts with patterns-sap-hana and will be uninstalled (which will affect a running cluster)
+    - unless: rpm -q SAPHanaSR-ScaleOut
```
arbulu89, line 18: I'm not sure about this change.
This looks a bug in our pattern or the package in any case, we shouldn't have this conflict

arbulu89 (changes requested):
Hi @yeoldegrove ,
Loved the changes about the ini files.
Anyway, I think that the changes on the monitoring configuration are not good, and they will lead to errors in some scenarios.
The same happens with the pattern installation. This 2nd thing looks a sensitive thing. We should raise this error to the pattern maintainer.

hana/packages.sls
yeoldegrove: After a discussion with the maintainer of the `patterns-sap-hana` pacakges, this seems to be an unreasonable conflict indeed at the moment.
Maybe a solution will come in the future but for now we have to do it this way.

PR #136, opened 2021-11-11
use multi-target Hook on HANA scale-out and use sudoers.d file

yeoldegrove:
Starting with version 0.180 the SAPHanaSR-ScaleOut package supports SAP HANA scale-out multi-target system replication. That means you can connect a third HANA site by system replication to either of the two HANA sites which are managed by the SUSE HA cluster.

More Details can be found here:
https://www.suse.com/c/sap-hana-scale-out-upgrade-details/
https://www.suse.com/c/sap-hana-scale-out-multi-target-upgrade/

To enable this feature, a different hook script needs to be used:
~~/usr/share/SAPHanaSR-ScaleOut/SAPHanaSR.py~~
/usr/share/SAPHanaSR-ScaleOut/SAPHanaSrMultiTarget.py

This PR enables the new Hook if it is available and also migrates old setups (hook and sudoers).

fixes https://github.com/SUSE/saphanabootstrap-formula/issues/135 
fixes https://github.com/SUSE/ha-sap-terraform-deployments/issues/787

### DIFF
--- a/_service
+++ b/_service
@@ -4,7 +4,7 @@
     <param name="scm">git</param>
     <param name="exclude">.git</param>
     <param name="filename">saphanabootstrap-formula</param>
-    <param name="versionformat">0.8.0+git.%ct.%h</param>
+    <param name="versionformat">0.8.1+git.%ct.%h</param>
     <param name="revision">%%VERSION%%</param>
   </service>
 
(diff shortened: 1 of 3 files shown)


melzer-b1 (approved):
Reviewed. Thanks Eike

fmherschel:
This fix completely makes sense in top-level-view (architecture view).
So far I did not have checked the code but using the multi-target instead of the "old" hook is completely ok.
The internal migration should NOT be done automatically (so no attributes should be changed here).

PR #144, opened 2022-07-25
make sudoers rules more secure

yeoldegrove:
This PR tackles https://bugzilla.suse.com/show_bug.cgi?id=1196642 which suggests:
- to get rid of the temporary unsecure sudoers file in `/tmp/` 
- do not use `*` in suders rules

The outcome of a discussion with the owners of https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-scaleOut-PerfOpt-15/index.html was, that it is OK to go this way (getting rid of `*`). The best practices have to be seen as a general guideline and might or might not suggest security best practices.

In addition to the changes named above, it introduces a template file for the SAPHanaSR sudoers. This way the templating is done in the template instead of the state and we can pass a [salt slot](https://docs.saltproject.io/en/latest/topics/slots/index.html) to evaluate the "srHookGen" at a very late stage (after the rpm dependencies are met).

Test deployments on azure and gcp and on scale-up and scale-out deployments have been successfully done.
A test package build is available here: https://build.opensuse.org/package/show/home:waldt:branches:network:ha-clustering:sap-deployments:devel/saphanabootstrap-formula

### DIFF
--- a/_service
+++ b/_service
@@ -4,7 +4,7 @@
     <param name="scm">git</param>
     <param name="exclude">.git</param>
     <param name="filename">saphanabootstrap-formula</param>
-    <param name="versionformat">0.10.1+git.%ct.%h</param>
+    <param name="versionformat">0.11.0+git.%ct.%h</param>
     <param name="revision">%%VERSION%%</param>
   </service>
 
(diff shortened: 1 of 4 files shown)


mr-stringer (approved):
LGTM

mgerstner:
While I'm not familiar with all the Salt / SAP details I believe the concerns about the sudo and tmp file handling I raised will be addressed with this change. Thanks for working on this!

PR #146, opened 2022-08-11
Gitfs compatibility

cschneemann:
Moved templates dir into hana states directory to be usable with gitfs.


cschneemann:
@yeoldegrove 
here now the PR for this formular. Did already bump the versions as you requested last times, hope I did it not too early ;)

yeoldegrove (changes requested):
@cschneemann Can't all references to `ftemplates` be removed? Also in the files section?

yeoldegrove (approved):
LGTM now. Also did a successfull test build and install.

PR #149, opened 2022-11-04
revert spec file changes

mr-stringer:
A previous change to the spec file prevented the build of RPMs for SLES.  This change has been reverted.

### DIFF
--- a/saphanabootstrap-formula.changes
+++ b/saphanabootstrap-formula.changes
@@ -1,5 +1,5 @@
 -------------------------------------------------------------------
-Friday Dec 04 11:27:52 UTC 2022 - Steve Stringer <steven.stringer@suse.com>
+Friday Nov 04 14:25:52 UTC 2022 - Steve Stringer <steven.stringer@suse.com>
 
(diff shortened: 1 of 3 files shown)


saphanabootstrap-formula.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Friday Dec 04 11:27:52 UTC 2022 - Steve Stringer <steven.stringer@suse.com>
```
angelabriel, line 2: Seems you computer is currently running in a parallel time line :-) My calendar tells me, we are still in November and not December.
Even that it is not a major problem it will be nice, if you can fix it.

angelabriel (changes requested):
Please fix date in changelog file.

saphanabootstrap-formula.changes
mr-stringer: Good spot, thanks.

Corrected and pushed.

angelabriel (approved):
Looks pretty good for me, many thanks.

PR #151, opened 2022-11-07
updated change file with correct date format

mr-stringer:
Used `osc vc` to write the change file to ensure correct format.


angelabriel (approved):
Looks good to me, thanks

PR #152, opened 2023-08-08
Add workaround to make "shaptools" available within Salt Bundle (bsc#1212695)

meaksh:
This PR simply introduces an small hack to the formula to make "shaptools" available inside current Salt Bundle environment, that is used for registered system in SUSE Manager.

The current extending mechanism for Salt Bundle doesn't make it easy to extend it with "shaptools" coming from an RPM package, so we did this small hack that works fine when using the current Salt Bundle.

Once we have a better mechanism for extending the Salt Bundle, then we should get rid of this workaround and go in favor of the new solution.

Tracks: https://github.com/SUSE/spacewalk/issues/21907

### DIFF
--- a/hana/packages.sls
+++ b/hana/packages.sls
@@ -42,13 +42,11 @@ python3-shaptools:
     - resolve_capabilities: true
 
 {# If venv-salt-minion is installed, then we make shaptools available on this environment #}
-{% if salt['pkg.version']('venv-salt-minion') %}
+{% if salt['pkg.version']('venv-salt-minion') and salt['file.file_exists']("/usr/bin/python3") %}
+{% set python_site_packages_path = salt["cmd.run"]("/usr/bin/python3 -c \"import sysconfig as s; print(s.get_paths().get('purelib'))\"") %}
+{% set salt_bundle_site_packages_path = salt["cmd.run"]("/usr/lib/venv-salt-minion/bin/python -c \"import sysconfig as s; print(s.get_paths().get('purelib'))\"") %}
 shaptools_available_in_salt_bundle:
   file.symlink:
-    - name: /usr/lib/venv-salt-minion/lib/python3.10/site-packages/shaptools
-    - target: /usr/lib/python3.6/site-packages/shaptools/
-    - onlyif:
-      - test -d /usr/lib/python3.6/site-packages/shaptools/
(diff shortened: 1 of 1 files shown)


hana/packages.sls
```diff
@@ -40,3 +40,15 @@ python3-shaptools:
         attempts: 3
         interval: 15
     - resolve_capabilities: true
+
+{# If venv-salt-minion is installed, then we make shaptools available on this environment #}
+{% if salt['pkg.version']('venv-salt-minion') %}
+shaptools_available_in_salt_bundle:
+  file.symlink:
+    - name: /usr/lib/venv-salt-minion/lib/python3.10/site-packages/shaptools
```
vzhestkov, line 48: Maybe slightly improve the solution with getting the proper `site-package` path with more flexible way:
```
python3 -c "import sysconfig as s; print(s.get_paths().get('purelib'))"
```
and for the bundle:
```
/usr/lib/venv-salt-minion/bin/python -c "import sysconfig as s; print(s.get_paths().get('purelib'))"
```

meaksh: Thanks for the suggestions Victor. I've adapted the SLS now.

agraul: IMO the hardcoded version is easier to read. `/usr/bin/python3` will always be 3.6 for SLE 15 and I don't think we'll change the location of the venv before we fix the extension.

vzhestkov: But we can bump the version of the python inside the bundle. Maybe hardcoded python version of the distro is ok, but in case of upgrading the python in the bundle we would have to adopt it again.

agraul: Before we change the Python version in the Bundle, we ought to fix the problem with installing shaptools inside of it properly. This PR is not the proper solution, it's an ugly hack that we should drop again as soon as we can.

meaksh: I see Alex's point, as mentioned, the final fix for this formula should be done once we enhance the extending mechanism for the Salt Bundle.

OTOH I think I would prefer to keep this hack as it is right now (by not hardcoding the python version) just in case we bump the Python version inside the Salt Bundle in the future (even when having already an enhanced bundle extending mechanism) but this formula has not been adapted/released yet, which will potentially cause another L3.

meaksh:
Closing this one in favor of https://github.com/SUSE/shaptools/pull/76

PR #153, opened 2023-11-06
add susChkSrv HA/DR provider and make providers configurable

yeoldegrove:
* add susChkSrv HA/DR provider
* make providers configurable

This will be a part of fixing https://github.com/SUSE/ha-sap-terraform-deployments/issues/910

### DIFF
--- a/hana/ha_cluster.sls
+++ b/hana/ha_cluster.sls
@@ -3,9 +3,6 @@
 
 {% if hana.scale_out %}
 {% set hook_path = '/usr/share/SAPHanaSR-ScaleOut' %}
-{% set sr_hook_multi_target = hook_path + '/SAPHanaSrMultiTarget.py' %}
-{% set sr_hook = hook_path + '/SAPHanaSR.py' %}
-{% set sustkover_hook = hook_path + '/susTkOver.py' %}
 
 remove_SAPHanaSR:
   pkg.removed:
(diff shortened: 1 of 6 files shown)


angelabriel (commented):
Looks good to me.
But I would prefer to switch the defaults for susTkOver and susChkSrv from enabled to disabled because both hooks are NOT mandatory. So the customer should actively enable the hooks so that he knows what is going on on his HANA cluster.

stefanotorresi:
I agree that the additional hooks should be disabled by default.

angelabriel (approved):
Looks good to me. Thanks for switching the default :smile:

hana/ha_cluster.sls
```diff
@@ -38,6 +32,24 @@ install_SAPHanaSR:
       - SAPHanaSR-doc
 {% endif %}
 
+{% set sr_hook_multi_target = hook_path + '/SAPHanaSrMultiTarget.py' %}
+{% set sr_hook = hook_path + '/SAPHanaSR.py' %}
+{% set sustkover_hook = hook_path + '/susTkOver.py' %}
+{% set suschksrv_hook = hook_path + '/susChkSrv.py' %}
+
+{% set sustkover_hook_enabled = False %}
```
arbulu89, line 40: This looks a bit strange.
Isn't any option to set a default value if the value doesn't exist?
I mean, i find it rare this `if` condition usage. It is a pretty basic feature
Something like:

```
{% set sustkover_hook_enabled = hana.ha_dr_sustkover_enabled|default(False) %}
```

arbulu89 (approved):
It looks good!
Only one comment so you can improve the yaml code

hana/ha_cluster.sls
yeoldegrove: Let me double check if the approach you mention (using `default()`) works now... My fist attempt was to use a Dict like `hana.ha_dr.sustkover.enabled` ...
And I had issues if one of the "levels" did not exist. As we do not do any yaml validation we have issues like this :(
Now that I use a flat variable structure, this could work... Will come back.

yeoldegrove: I did quite some testing now passing `default()`, with several inputs and it works for me.
Pushed the changes you suggested.

